PR #1, opened 2019-07-02
Implement methods for amqp header exchange

izzymoriguchi:
### Major changes
- Created the following classes
    - `AmqpConfiguration`
    - `AmqpController`
    - `AmqpControllerFactorySpi`
    - `AmqpElektron`
    - `AmqpNukleus`
    - `AmqpNukleusFactorySpi`
    - `AmqpClientFactoryBuilder` (Not implemented yet)
    - `AmqpServerFactory`
    - `AmqpServerFactoryBuilder`
    - `AmqpFrameFW`
    - `AmqpHeaderFW` (Not implemented yet)

- `AmqpServerFactory`
	- Implemented methods that the header exchange test case requires to pass

### DIFF
--- a/src/main/java/org/reaktivity/nukleus/amqp/internal/AmqpConfiguration.java
+++ b/src/main/java/org/reaktivity/nukleus/amqp/internal/AmqpConfiguration.java
@@ -28,7 +28,7 @@ public class AmqpConfiguration extends Configuration
     static
     {
         final ConfigurationDef config = new ConfigurationDef("nukleus.amqp");
-        AMQP_CONTAINER_ID = config.property("container.id", "container-id:1");
+        AMQP_CONTAINER_ID = config.property("container.id");
         AMQP_CHANNEL_MAX = config.property("channel.max", 65535);
         AMQP_MAX_FRAME_SIZE = config.property("max.frame.size", 4294967295L);
         AMQP_CONFIG = config;
--- a/src/main/java/org/reaktivity/nukleus/amqp/internal/AmqpConfiguration.java
+++ b/src/main/java/org/reaktivity/nukleus/amqp/internal/AmqpConfiguration.java
@@ -1,0 +1,57 @@
+/**
+ * Copyright 2016-2019 The Reaktivity Project
+ *
+ * The Reaktivity Project licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package org.reaktivity.nukleus.amqp.internal;
+
+import org.reaktivity.nukleus.Configuration;
+
+public class AmqpConfiguration extends Configuration
+{
+    public static final PropertyDef<String> AMQP_CONTAINER_ID;
+    public static final IntPropertyDef AMQP_CHANNEL_MAX;
+    public static final LongPropertyDef AMQP_MAX_FRAME_SIZE;
+
+    private static final ConfigurationDef AMQP_CONFIG;
+
+    static
+    {
+        final ConfigurationDef config = new ConfigurationDef("nukleus.amqp");
+        AMQP_CONTAINER_ID = config.property("container.id", "container-id:1");
+        AMQP_CHANNEL_MAX = config.property("channel.max", 65535);
+        AMQP_MAX_FRAME_SIZE = config.property("max.frame.size", 4294967295L);
+        AMQP_CONFIG = config;
+    }
+
+    public AmqpConfiguration(
+        Configuration config)
+    {
+        super(AMQP_CONFIG, config);
+    }
+
+    public String containerId()
+    {
+        return AMQP_CONTAINER_ID.get(this);
+    }
+
+    public int channelMax()
+    {
+        return AMQP_CHANNEL_MAX.getAsInt(this);
+    }
+
+    public long maxFrameSize()
+    {
+        return AMQP_MAX_FRAME_SIZE.getAsLong(this);
+    }
+}
--- a/src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
+++ b/src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
@@ -23,8 +23,10 @@ import java.util.function.LongUnaryOperator;
 import org.agrona.DirectBuffer;
 import org.agrona.MutableDirectBuffer;
 import org.agrona.collections.Long2ObjectHashMap;
+import org.agrona.concurrent.UnsafeBuffer;
 import org.reaktivity.nukleus.amqp.internal.types.codec.AmqpFrameFW;
 import org.reaktivity.nukleus.amqp.internal.types.codec.AmqpHeaderFW;
+import org.reaktivity.nukleus.amqp.internal.types.codec.AmqpOpenFrameFW;
 import org.reaktivity.nukleus.amqp.internal.types.stream.*;
 import org.reaktivity.nukleus.buffer.BufferPool;
 import org.reaktivity.nukleus.function.MessageConsumer;
@@ -73,6 +75,7 @@ public final class AmqpServerFactory implements StreamFactory
 
     private final RouteManager router;
     private final MutableDirectBuffer writeBuffer;
+    private final MutableDirectBuffer encodeBuffer;
     private final LongUnaryOperator supplyInitialId;
     private final LongUnaryOperator supplyReplyId;
     private final LongSupplier supplyTraceId;
@@ -101,6 +104,7 @@ public final class AmqpServerFactory implements StreamFactory
         this.correlations = new Long2ObjectHashMap<>();
         this.sessions = new Long2ObjectHashMap<>();
         this.wrapRoute = this::wrapRoute;
+        this.encodeBuffer = new UnsafeBuffer(new byte[writeBuffer.capacity()]);
     }
 
     @Override
@@ -133,6 +137,7 @@ public final class AmqpServerFactory implements StreamFactory
     {
         final long routeId = begin.routeId();
         final long initialId = begin.streamId();
+        final long replyId = supplyReplyId.applyAsLong(initialId);
 
         final MessagePredicate filter = (t, b, o, l) -> true;
         final RouteFW route = router.resolve(routeId, begin.authorization(), filter, this::wrapRoute);
@@ -140,8 +145,8 @@ public final class AmqpServerFactory implements StreamFactory
 
         if (route != null)
         {
-            final AmqpServerConnection connection = new AmqpServerConnection(sender, routeId, initialId);
-            correlations.put(supplyReplyId.applyAsLong(initialId), connection);
+            final AmqpServerConnection connection = new AmqpServerConnection(sender, routeId, initialId, replyId);
+            correlations.put(replyId, connection);
             newStream = connection::onNetwork;
         }
         return newStream;
@@ -152,12 +157,8 @@ public final class AmqpServerFactory implements StreamFactory
         final MessageConsumer sender)
     {
         final long replyId = begin.streamId();
-        final AmqpServerConnection connection = correlations.remove(replyId);
+        // TODO: Application stream
         MessageConsumer newStream = null;
-        if (connection != null)
-        {
-            newStream = connection::onNetwork;
-        }
         return newStream;
     }
 
@@ -184,16 +185,19 @@ public final class AmqpServerFactory implements StreamFactory
 
         private long decodeTraceId;
         private DecoderState decodeState;
+        private int bufferSlot = BufferPool.NO_SLOT;
+        private int bufferSlotOffset;
 
         private AmqpServerConnection(
             MessageConsumer receiver,
             long routeId,
-            long initialId)
+            long initialId,
+            long replyId)
         {
             this.receiver = receiver;
             this.routeId = routeId;
             this.initialId = initialId;
-            this.replyId = supplyReplyId.applyAsLong(initialId);
+            this.replyId = replyId;
             this.decodeState = this::decodeHeader;
         }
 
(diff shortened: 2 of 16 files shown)


src/main/java/org/reaktivity/nukleus/amqp/internal/AmqpConfiguration.java
```diff
@@ -0,0 +1,57 @@
...
+    public static final PropertyDef<String> AMQP_CONTAINER_ID;
+    public static final IntPropertyDef AMQP_CHANNEL_MAX;
+    public static final LongPropertyDef AMQP_MAX_FRAME_SIZE;
+
+    private static final ConfigurationDef AMQP_CONFIG;
+
+    static
+    {
+        final ConfigurationDef config = new ConfigurationDef("nukleus.amqp");
+        AMQP_CONTAINER_ID = config.property("container.id", "container-id:1");
```
jfallows: Do we need to default the `containerId`?
Perhaps we could instead use the local address name itself as a default, such as `amqp#0`?

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -0,0 +1,512 @@
...
+        final long initialId = begin.streamId();
+
+        final MessagePredicate filter = (t, b, o, l) -> true;
+        final RouteFW route = router.resolve(routeId, begin.authorization(), filter, this::wrapRoute);
+        MessageConsumer newStream = null;
+
+        if (route != null)
+        {
+            final AmqpServerConnection connection = new AmqpServerConnection(sender, routeId, initialId);
+            correlations.put(supplyReplyId.applyAsLong(initialId), connection);
```
jfallows: Hoist `replyId` into separate variable for readability and pass to `AmqpServerConnection`, no need to recompute `replyId` in the constructor.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -0,0 +1,512 @@
...
+    private MessageConsumer newReplyStream(
+        final BeginFW begin,
+        final MessageConsumer sender)
+    {
+        final long replyId = begin.streamId();
+        final AmqpServerConnection connection = correlations.remove(replyId);
+        MessageConsumer newStream = null;
+        if (connection != null)
+        {
+            newStream = connection::onNetwork;
```
jfallows: AMQP reply streams for the server will be application streams, not network streams, so this should be left as a `TODO` with `newStream` remaining `null` for now.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -0,0 +1,512 @@
...
+                decodeTraceId = data.trace();
+                final OctetsFW payload = data.payload();
+                final DirectBuffer buffer = payload.buffer();
+
+                int offset = payload.offset();
+                int length = payload.sizeof();
+
+                while (length > 0)
+                {
+                    int consumed = decodeState.decode(buffer, offset, length);
```
jfallows: We also need to detect no change in `decodeState` and break the loop, even if length is still non-zero.
This represents the case where we have insufficient data to completely decode a frame.
Any unprocessed payload then needs to be stored in a buffer slot for the stream, and subsequent `DATA` frames need to have their payload appended at the end of any data in the buffer slot, so that the current decode state can try again when more payload arrives from the network.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -0,0 +1,512 @@
...
+
+        private void onSignal(
+            SignalFW signal)
+        {
+            // TODO
+        }
+
+        private void onAmqpHeader(AmqpHeaderFW header)
+        {
+            doAmqpHeader(header.buffer(), header.offset(), header.limit());
```
jfallows: We need some validation here to ensure we are speaking a consistent version of `AMQP 1.0`.
If the initial bytes do not represent a valid header exchange, then the connection should be rejected with appropriate error code. Likely we need to add this as a negative spec test in `nukleus-amqp.spec`.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -0,0 +1,512 @@
...
+        {
+            // TODO
+        }
+
+        private int decodeHeader(
+            final DirectBuffer buffer,
+            final int offset,
+            final int length)
+        {
+            final AmqpHeaderFW amqpHeader = amqpHeaderRO.wrap(buffer, offset, offset + length);
```
jfallows: Use `tryWrap` instead of `wrap` to avoid exception and get `null` return value instead when wrapped data is insufficient.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -0,0 +1,512 @@
...
+            onAmqpHeader(amqpHeader);
+            decodeState = this::decodeFrame;
```
jfallows: Change the `decodeState` in `onAmqpHeader`, since the next state may vary depending on validity of the AMQP protocol header bytes offered by the client.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -0,0 +1,512 @@
...
+                    onAmqpDetach(amqpFrame);
+                    break;
+                case 0x07:
+                    onAmqpEnd(amqpFrame);
+                    break;
+                case 0x08:
+                    onAmqpClose(amqpFrame);
+                    break;
+            }
+            return amqpFrame.sizeof();
```
jfallows: Does `amqpFrame` wrap the entire frame, or just the generic frame header?
That will determine whether or not this return value is correct since you want the decode loop to skip this many bytes to advance to the beginning of the next AMQP frame when progressing through the network stream.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -0,0 +1,512 @@
...
+    @FunctionalInterface
+    private interface DecoderState
+    {
+        int decode(DirectBuffer buffer, int offset, int length);
+    }
+
+    // TODO
+    private final class AmqpServerSession
+    {
+        private final Long2ObjectHashMap<AmqpServerSession> links;
```
jfallows: Should `links` be parametric in `AmqpServerStream`?

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactoryBuilder.java
```diff
@@ -0,0 +1,137 @@
...
+import org.reaktivity.nukleus.amqp.internal.AmqpConfiguration;
+import org.reaktivity.nukleus.buffer.BufferPool;
+import org.reaktivity.nukleus.route.RouteManager;
+import org.reaktivity.nukleus.stream.StreamFactory;
+import org.reaktivity.nukleus.stream.StreamFactoryBuilder;
+
+public final class AmqpServerFactoryBuilder implements StreamFactoryBuilder
+{
+    private final AmqpConfiguration config;
+//    private final Long2ObjectHashMap<Correlation<?>> correlations;
```
jfallows: Remove commented code.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactoryBuilder.java
```diff
@@ -0,0 +1,137 @@
...
+
+import java.util.function.IntUnaryOperator;
+import java.util.function.LongFunction;
+import java.util.function.LongSupplier;
+import java.util.function.LongUnaryOperator;
+import java.util.function.Supplier;
+import java.util.function.ToIntFunction;
+
+import org.agrona.MutableDirectBuffer;
+//import org.agrona.collections.Long2ObjectHashMap;
```
jfallows: Remove commented code.

jfallows (changes requested):
Excellent start @izzymoriguchi !

src/main/reaktivity/protocol.idl
```diff
@@ -0,0 +1,780 @@
...
+            CLOSE (0x18L),
+            ERROR (0x1dL),
+            HEADER (0x70L),
+            DELIVERY_ANNOTATIONS (0x71L),
+            MESSAGE_ANNOTATIONS (0x72L),
+            PROPERTIES (0x73L),
+            APPLICATION_PROPERTIES (0x74L),
+            DATA (0x75L),
+            AMQP_SEQUENCE (0x76L),
+            AMQP_VALUE (0x77L),
```
jfallows: No need for `AMQP_` prefix, right?

src/main/reaktivity/protocol.idl
```diff
@@ -0,0 +1,780 @@
...
+            SASL_INIT (0x41L),
+            SASL_CHALLENGE (0x42L),
+            SASL_RESPONSE (0x43L),
+            SASL_OUTCOME (0x44L),
+        }
+
+        // 1.6 primitive type definitions
+        variant AmqpNull switch (AmqpType)
+        {
+            case NULL: 0;
```
jfallows: should this be `null` instead of `0` ?

src/main/reaktivity/protocol.idl
```diff
@@ -0,0 +1,780 @@
...
+        }
+
+        variant AmqpBoolean switch (AmqpType) of uint8
+        {
+            case BOOLEAN: uint8;
+            case TRUE: 1;
+            case FALSE: 0;
+        }
```
jfallows: Do we need first class `bool` type?

src/main/reaktivity/protocol.idl
```diff
@@ -0,0 +1,780 @@
...
+            AmqpArray<AmqpIETFLanguageTag> outgoingLocales;
+            AmqpArray<AmqpIETFLanguageTag> incomingLocales;
+            AmqpArray<AmqpSymbol> offeredCapabilities;
+            AmqpArray<AmqpSymbol> desiredCapabilities;
+            AmqpFields<AmqpValue> properties;
+        }
+
+        list AmqpBegin (AmqpList)
+        {
+            AmqpUShort remoteChannel;
```
jfallows: what does it mean for `remoteChannel` not to be `required` when subsequent fields are `required`?

src/main/reaktivity/protocol.idl
```diff
@@ -0,0 +1,780 @@
...
+        typedef AmqpUInt as AmqpMilliseconds;
+        typedef AmqpBinary as AmqpDeliveryTag;
+        typedef AmqpSequenceNo as AmqpDeliveryNumber;
+        typedef AmqpSequenceNo as AmqpTransferNumber;
+        typedef AmqpUInt as AmqpSequenceNo;
+        typedef AmqpUInt as AmqpMessageFormat;
+        typedef AmqpSymbol as AmqpIETFLanguageTag;
+        typedef AmqpMap<AmqpSymbol, V> as AmqpFields<V>;
+
+        list ErrorList (AmqpList)
```
jfallows: Rename to `AmqpErrorList` ?

src/main/reaktivity/protocol.idl
```diff
@@ -0,0 +1,780 @@
...
+            SESSION_HANDLE_IN_USE ("amqp:session:handle-in-use"),
+            SESSION_UNATTACHED_HANDLE ("amqp:session:unattached-handle"),
+            LINK_DETACH_FORCED ("amqp:link:detach-forced"),
+            LINK_TRANSFER_LIMIT_EXCEEDED ("amqp:link:transfer-limit-exceeded"),
+            LINK_MESSAGE_SIZE_EXCEEDED ("amqp:link:message-size-exceeded"),
+            LINK_REDIRECT ("amqp:link:redirect"),
+            LINK_STOLEN ("amqp:link:stolen"),
+            TRANSACTION_UNKNOWN_ID ("amqp:transaction:unknown-id"),
+            TRANSACTION_ROLLBACK ("amqp:transaction:rollback"),
+            TRANSACTION_TIMEOUT ("amqp:transaction:timeout")
```
jfallows: Suggest putting a newline between groups of errors for increased readability only.

src/main/reaktivity/protocol.idl
```diff
@@ -0,0 +1,780 @@
...
+            case LIST1:
+            case LIST4:
+                AmqpList;
+            case MAP1:
+            case MAP4:
+                AmqpMap;
+            case ARRAY1:
+            case ARRAY4:
+                AmqpArray;
+        }
```
jfallows: Do we still need `AmqpValue` ?

jfallows (changes requested):

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -0,0 +1,613 @@
...
+        }
+
+        private void onAmqpHeader(
+            AmqpProtocolHeaderFW header)
+        {
+            if (header != null)
+            {
+                if (isAmqpHeaderValid(header))
+                {
+                    doAmqpHeader(header.buffer(), header.offset(), header.limit());
```
jfallows: Suggest switching this to receive a major version, a minor version and a revision version, then construct the AMQP protocol header in `doAmqpHeader` rather than echoing back the same `header` as was received.

You'll certainly need an approach that doesn't rely on the input when handling the invalid protocol header case.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -0,0 +1,613 @@
...
+    private RouteFW wrapRoute(
+        int msgTypeId,
+        DirectBuffer buffer,
+        int index,
+        int length)
+    {
+        return routeRO.wrap(buffer, index, index + length);
+    }
+
+    private final class AmqpServerConnection
```
jfallows: Suggest we rename this to `AmqpServer` instead.

jfallows (changes requested):

jfallows (dismissed):

jfallows (approved):

PR #10, opened 2020-03-16
Implement amqp subscribe

izzymoriguchi:


src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -170,150 +225,507 @@ private MessageConsumer newReplyStream(
...
+            int offsetOpenFrame = amqpFrameHeaderRW.limit();
+
+            final String8FW containerIdRO = new String8FW.Builder()
+                .wrap(writeBuffer, 0, writeBuffer.capacity())
+                .set(containerId, UTF_8)
+                .build();
+
+            final AmqpOpenFW open = amqpOpenRW
+                .wrap(frameBuffer, offsetOpenFrame, frameBuffer.capacity())
+                .containerId(containerIdRO)
```
izzymoriguchi: In general, I am not completely sure which fields I should fill in (out of all the non-required fields). 
1. Do I encode only the present fields that were sent by the client? 
2. If so, should we have convenience methods that check whether each field is present in `AmqpOpenFW`?  (Currently accessors of each field in `AmqpOpenFW` gives `assertion error` if we call fields that are not present.)

For `offered-capabilities` and `desired-capabilities` fields in `list AmqpOpen`, it seems like if `desired-capabilities` was sent by a client, then we send back `offered-capabilities`.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -170,150 +225,507 @@ private MessageConsumer newReplyStream(
...
-            final EndFW end = endRW.wrap(writeBuffer, 0, writeBuffer.capacity())
-                .routeId(routeId)
-                .streamId(replyId)
-                .traceId(supplyTraceId.getAsLong())
+            int offsetBeginFrame = amqpFrameHeaderRW.limit();
+
+            final AmqpBeginFW begin = amqpBeginRW
+                .wrap(frameBuffer, offsetBeginFrame, frameBuffer.capacity())
+                .remoteChannel(1)
+                .nextOutgoingId(1)
```
izzymoriguchi: For `remoteChannel` and `nextOutgoingId`, I am still trying to figure out what values to encode for these values.

izzymoriguchi:
### Possible remaining `maven-plugin` language features to support for `AmqpAttachFW`,  `AmqpTransferFW`, `AmqpFlowFW`, and `AmqpCloseFW`
- `enum` with `variant` 
  - e.g. `enum AmqpDescribedType (AmqpDescriptor)`
-  `list` : allow default value of `enum`  and `variant` 
    -  e.g. `list AmqpAttach `→  `AmqpSenderSettleMode sndSettleMode = MIXED; `
    - e.g. `list AmqpAttach` → `AmqpBoolean incompleteUnsettled = FALSE; `
- OPTIONLAL? `list` : have methods to check whether field is present (not null)
- `variant` with `variant` as a type of field (instead of uint64, or uint8 primitive types) 
  - e.g. v`ariant AmqpDescriptor `→ `case DESCRIBED: AmqpULong; `
- `union` that takes in `enum`
  - e.g.  `union AmqpOutcome switch (AmqpDescribedType) `

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -48,6 +84,8 @@
 
 public final class AmqpServerFactory implements StreamFactory
 {
+    private static final OctetsFW EMPTY_OCTETS = new OctetsFW().wrap(new UnsafeBuffer(new byte[0]), 0, 0);
```
jfallows: Use no-args constructor for `UnsafeBuffer` to create empty buffer.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -56,63 +94,121 @@
...
         this.supplyInitialId = requireNonNull(supplyInitialId);
         this.supplyReplyId = requireNonNull(supplyReplyId);
+        this.supplyBudgetId = requireNonNull(supplyBudgetId);
         this.supplyTraceId = requireNonNull(supplyTraceId);
         this.correlations = new Long2ObjectHashMap<>();
-        this.sessions = new Long2ObjectHashMap<>();
-        this.wrapRoute = this::wrapRoute;
-        this.encodeBuffer = new UnsafeBuffer(new byte[writeBuffer.capacity()]);
+        this.amqpTypeId = supplyTypeId.applyAsInt(AmqpNukleus.NAME);
+        this.defaultContainerId = config.containerId();
```
jfallows: This is just `containerId`, right?

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -56,63 +94,121 @@
...
+    private final MutableDirectBuffer terminusListBuffer;
     private final LongUnaryOperator supplyInitialId;
     private final LongUnaryOperator supplyReplyId;
     private final LongSupplier supplyTraceId;
+    private final LongSupplier supplyBudgetId;
 
-    private final Long2ObjectHashMap<AmqpServer> correlations;
-    private final Long2ObjectHashMap<AmqpServerSession> sessions;
-    private final MessageFunction<RouteFW> wrapRoute;
+    private final Long2ObjectHashMap<AmqpServer.AmqpServerStream> correlations;
```
jfallows: Change type of `correlations` to `Long2ObjectHashMap<MessageConsumer>` as it will make `newReplyStream` easier to implement.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -165,155 +263,714 @@ private MessageConsumer newReplyStream(
         final MessageConsumer sender)
     {
         final long replyId = begin.streamId();
-        // TODO: Application stream
+        final AmqpServer.AmqpServerStream reply = correlations.remove(replyId);
```
jfallows: After changing type of `correlations` as noted above, this will simplify to:
```
final long replyId = begin.streamId();
return correlations.remove(replyId);
```

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -144,17 +240,19 @@ private MessageConsumer newInitialStream(
...
         if (route != null)
         {
-            final AmqpServer connection = new AmqpServer(sender, routeId, initialId, replyId);
-            correlations.put(replyId, connection);
+            final long initialId = begin.streamId();
+            final long affinity = begin.affinity();
+            final long replyId = supplyReplyId.applyAsLong(initialId);
+            final long budgetId = supplyBudgetId.getAsLong();
+
+            final AmqpServer connection = new AmqpServer(sender, routeId, initialId, replyId, affinity, budgetId);
```
jfallows:
```
newStream = new AmqpServer(sender, routeId, initialId, replyId, affinity, budgetId)::onNetwork;
```

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -165,155 +263,714 @@ private MessageConsumer newReplyStream(
...
+    {
+        int progress = offset;
+        final AmqpFrameHeaderFW frameHeader = amqpFrameHeaderRO.tryWrap(buffer, offset, limit);
+
+        if (frameHeader != null)
+        {
+            final AmqpFrameType frameType = AmqpFrameType.valueOf(frameHeader.performative().descriptorId().getAsUint8());
+            final AmqpServerDecoder decoder = decodersByFrameType.getOrDefault(frameType, decodeUnknownType);
+
+            if (limit - frameHeader.limit() >= 0)
```
jfallows: Why do we need this condition?

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -165,155 +263,714 @@ private MessageConsumer newReplyStream(
...
+            AmqpFrameHeaderFW originalFrameHeader = amqpFrameHeaderRW.build();
+
+            final AmqpFrameHeaderFW updatedFrameHeader =
+                amqpFrameHeaderRW.wrap(frameBuffer, 0, frameBuffer.capacity())
+                    .size(originalFrameHeader.sizeof() + attach.sizeof())
+                    .doff(originalFrameHeader.doff())
+                    .type(originalFrameHeader.type())
+                    .channel(originalFrameHeader.channel())
+                    .performative(b -> b.domainId(d -> d.set(originalFrameHeader.performative().domainId()))
+                        .descriptorId(d2 -> d2.set(originalFrameHeader.performative().descriptorId().get())))
```
jfallows: I have a lot of this style code in `kafka` nukleus as well.

Do you think we could modify the maven plugin to support typed non-primitive setters?
```
.channel(originalFrameHeader.channel())
.performative(originalFrameHeader.performative())
```
The resulting code above would be much simpler and more efficient.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -326,163 +983,392 @@ private void onNetwork(
...
+                cleanupDecodeSlotIfNecessary();
+            }
+        }
+
+        private void onDecodeHeader(
+            long traceId,
+            long authorization,
+            AmqpProtocolHeaderFW header)
+        {
+            if (isAmqpHeaderValid(header))
```
jfallows, line 1461: What if the header is not valid?

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -326,163 +983,392 @@ private void onNetwork(
...
+        {
+            doEncodeFrameHeader(frameHeader.doff(), frameHeader.type(), frameHeader.channel(), frameHeader.performative());
+        }
+
+        private void onDecodeOpen(
+            long traceId,
+            long authorization,
+            AmqpOpenFW open)
+        {
+            containerId = open.containerId().asString();
```
jfallows: Do we still need to retain this?
I think we were going to remove `containerId` from `AmqpBeginEx`.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -326,163 +983,392 @@ private void onNetwork(
...
+        {
+            if (begin.hasRemoteChannel() || sessions.containsKey(channelId))
+            {
+                onDecodeError(traceId, authorization, NOT_ALLOWED);
+            }
+            else
+            {
+                channelId++;
+                nextTransferId++;
+                AmqpSession session = sessions.computeIfAbsent(channelId, s -> new AmqpSession());
```
jfallows: Change `AmqpSession` constructor to take `channelId` parameter (only), then use:
```
AmqpSession session = sessions.computeIfAbsent(channelId, AmqpSession::new);
```

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -326,163 +983,392 @@ private void onNetwork(
...
+            long authorization,
+            AmqpAttachFW attach)
+        {
+            if (!sessions.containsKey(channelId))
+            {
+                onDecodeError(traceId, authorization, NOT_ALLOWED);
+            }
+            else
+            {
+                sessions.get(channelId).onDecodeAttach(traceId, authorization, attach);
```
jfallows: Move `session.get(...)` to top, then test for `null` instead of doing both `contains` and `get` in the success case.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -326,163 +983,392 @@ private void onNetwork(
...
+        }
+
+        private void onDecodeClose(
+            long traceId,
+            long authorization,
+            AmqpCloseFW close)
+        {
+            if (close.fieldCount() == 0)
+            {
+                doEncodeClose(traceId, authorization, null);
```
jfallows, line 1554: We need to implicitly tear down any active sessions.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -501,148 +1387,699 @@ private boolean isAmqpHeaderValid(
...
         {
-            // TODO
+            private AmqpSession()
+            {
+                // TODO
+            }
+
+            private void onDecodeBegin(
+                long traceId,
+                long authorization)
```
jfallows: We need the `AmqpBeginFW` as a parameter, then assert the `channelId` is as expected from constructor.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -501,148 +1387,699 @@ private boolean isAmqpHeaderValid(
...
+                long traceId,
+                long authorization,
+                AmqpAttachFW attach)
+            {
+                final int linkNameKey = linkNameKey(attach.name().asString());
+                if (links.containsKey(linkNameKey))
+                {
+                    onDecodeError(traceId, authorization, NOT_ALLOWED);
+                }
+                AmqpServerStream link = links.computeIfAbsent(linkNameKey, l -> new AmqpServerStream());
```
jfallows: This is just `put` now, thanks to guard condition above, right?
```
AmqpServerStream link = new AmqpServerStream();
AmqpServerStream oldLink = links.put(linkNameKey, link);
assert oldLink == null;
```

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -501,148 +1387,699 @@ private boolean isAmqpHeaderValid(
...
+            {
+                doEncodeBegin(traceId, authorization);
+            }
+
+            private void onDecodeAttach(
+                long traceId,
+                long authorization,
+                AmqpAttachFW attach)
+            {
+                final int linkNameKey = linkNameKey(attach.name().asString());
```
jfallows: Rename to `linkKey`.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -501,148 +1387,699 @@ private boolean isAmqpHeaderValid(
...
+            private int initialSlot = NO_SLOT;
+            private int initialSlotOffset;
+            private long initialSlotTraceId;
+
+            private int state;
+            private Set<AmqpRole> roles;
+
+            AmqpServerStream()
+            {
+                this.roles = EnumSet.noneOf(AmqpRole.class);
```
jfallows: Excellent use of `EnumSet` 👍

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -501,148 +1387,699 @@ private boolean isAmqpHeaderValid(
...
+            }
+
+            private void onDecodeAttach(
+                long traceId,
+                long authorization,
+                AmqpAttachFW attach)
+            {
+                linkName = attach.name().asString();
+                handle = attach.handle();
```
jfallows: Use `this.` for assignments like `this.linkName` and `this.handle` to make it clear we're updating instance state.
Also `linkName` can just be `name`.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -501,148 +1387,699 @@ private boolean isAmqpHeaderValid(
...
+                state = AmqpState.closeInitial(state);
+                cleanupInitialSlotIfNecessary();
+
+                if (AmqpState.closed(state))
+                {
+                    // TODO
+                }
+            }
+
+            private void onApplicationInitial(
```
jfallows: Rename to `onApplication`, as it now covers both initial and reply application frames.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -501,148 +1387,699 @@ private boolean isAmqpHeaderValid(
...
+                // TODO
+
+                return false;
+            }
         }
+    }
 
-        private void onAmqpEnd(
-            AmqpFrameFW end)
+    private static final class AmqpState
```
jfallows: Move this to a separate top-level class.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -501,148 +1387,699 @@ private boolean isAmqpHeaderValid(
...
+    {
+        private static final int INITIAL_OPENING = 0x10;
+        private static final int INITIAL_OPENED = 0x20;
+        private static final int INITIAL_CLOSING = 0x40;
+        private static final int INITIAL_CLOSED = 0x80;
+        private static final int REPLY_OPENED = 0x01;
```
jfallows: You might need to add `REPLY_OPENING` state, like `KafkaState` and others.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -165,155 +263,714 @@ private MessageConsumer newReplyStream(
...
+            AmqpRole role,
+            AmqpSenderSettleMode senderSettleMode,
+            AmqpReceiverSettleMode receiverSettleMode,
+            String address)
         {
-            if (initialCredit > 0)
+            int offsetAttachFrame = amqpFrameHeaderRW.limit();
+
+            AmqpAttachFW.Builder attachRW = amqpAttachRW.wrap(frameBuffer, offsetAttachFrame, frameBuffer.capacity())
+                .name(asStringFW(linkName))
```
jfallows: You can just pass `linkName` directly here without `asStringFW`, right?
If all similar method call sites can be removed, then please also remove `asStringFW` method.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -165,155 +263,714 @@ private MessageConsumer newReplyStream(
...
+            }
+            else if (role == AmqpRole.RECEIVER)
             {
-                initialBudget += initialCredit;
+                AmqpTargetListFW targetList = new AmqpTargetListFW.Builder()
+                    .wrap(terminusListBuffer, 0, terminusListBuffer.capacity())
+                    .address(asStringFW(address))
+                    .build();
+                attachRW.target(b -> b.targetList(targetList));
+            }
```
jfallows, line 874: `switch` on `role`

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -501,148 +1387,699 @@ private boolean isAmqpHeaderValid(
...
+                if (role == AmqpRole.RECEIVER)
+                {
+                    route = resolveTarget(routeId, authorization, hasSourceAddress ?
+                        attach.source().sourceList().address().asString() : "", AmqpRole.RECEIVER);
+                }
+                else
+                {
+                    route = resolveTarget(routeId, authorization, hasTargetAddress ?
+                        attach.target().targetList().address().asString() : "", AmqpRole.SENDER);
+                }
```
jfallows: `switch` on `role`

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -501,148 +1387,699 @@ private boolean isAmqpHeaderValid(
...
+
+                    switch (amqpBeginEx.role().get())
+                    {
+                    case RECEIVER:
+                        amqpRole = AmqpRole.RECEIVER;
+                        break;
+                    case SENDER:
+                        amqpRole = AmqpRole.SENDER;
+                        break;
+                    }
```
jfallows, line 1992: Use `AmqpTypeUtil` instead of inlining here.

src/test/java/org/reaktivity/nukleus/amqp/internal/streams/server/AmqpServerIT.java
```diff
@@ -46,7 +46,7 @@
         .counterValuesBufferCapacity(8192)
         .nukleus("amqp"::equals)
         .affinityMask("target#0", EXTERNAL_AFFINITY_MASK)
-        .configure(AMQP_CONTAINER_ID.name(), "localhost")
+        .configure(AMQP_CONTAINER_ID.name(), "server")
```
jfallows: Can remove `.name()`, right?

jfallows (changes requested):

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
izzymoriguchi: Right, we don't need this.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
izzymoriguchi: For `list` `Builder`, if a field is `variant`, the setter's argument should be the `variant`'s `oftype` type. And since the type of the field `name` is `AmqpString` (which is `variant`), we can only set as `StringFW` right now.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
izzymoriguchi: The possible design for the non-premitive setters for `struct` builder:
```
// AmqpFrameHeaderFW.Builder
public Builder performative(
    PerformativeFW value)
{
    assert lastFieldSet == INDEX_PERFORMATIVE - 1;
    PerformativeFW.Builder performativeRW = this.performativeRW.wrap(buffer(), limit(), maxLimit());
    performativeRW.fields(value.buffer(), value.offset(), value.sizeof());
    limit(performativeRW.build().limit());
    lastFieldSet = INDEX_PERFORMATIVE;
    return this;
}
```
```
// PerformativeFW.Builder
public Builder fields(
    DirectBuffer buffer,
    int index,
    int length)
{
    int newLimit = limit() + length;
    checkLimit(newLimit, maxLimit());
    buffer().putBytes(limit(), buffer, index, length);
    limit(newLimit);
    lastFieldSet = INDEX_DESCRIPTOR_ID;
    return this;
}
```

So we also need to add additional `fields` method in `struct` builder. 

Should I implement these in `nukleus-maven-plugin`?

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
izzymoriguchi: Added an `else` statement with `onDecodeError` .

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
izzymoriguchi: Discussed and resolved on slack.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
izzymoriguchi: Discussed and resolved on Slack. (Use a StringFW shared flyweight instance and populate the string)

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
izzymoriguchi: We do not need to retain this `containerId` anymore. I removed `containerId` from `AmqpBeginEx` in `nukleus-amqp.spec` and `command-log`.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
izzymoriguchi: According to the `AMQP` spec, _If a session is locally initiated, the remote-channel MUST NOT be set_. So when we receive `AmqpBeginFW` `begin` here, the `channelId` is null.

jfallows: Ah yes, thanks for the reminder. :-)

src/main/java/org/reaktivity/nukleus/amqp/internal/AmqpConfiguration.java
```diff
@@ -22,15 +22,18 @@
...
 
     static
     {
         final ConfigurationDef config = new ConfigurationDef("nukleus.amqp");
-        AMQP_CONTAINER_ID = config.property("container.id");
+        AMQP_CONTAINER_ID = config.property("container.id", "reaktivity");
         AMQP_CHANNEL_MAX = config.property("channel.max", 65535);
         AMQP_MAX_FRAME_SIZE = config.property("max.frame.size", 4294967295L);
+        AMQP_OUTGOING_WINDOW = config.property("outgoing.window", Integer.MAX_VALUE);
+        AMQP_INIT_DEVIVERY_COUNT = config.property("initial.delivery.count", 0L);
```
jfallows: Please spell out `INITIAL` instead of abbreviating as `INIT` to follow the naming convention consistently.

src/main/reaktivity/protocol.idl
```diff
@@ -294,238 +182,483 @@ scope protocol
...
+            AmqpULong descriptorId; // 0x53 0xXX
+        }
+
+        struct AmqpFrameHeader
+        {
+            uint32 size;
+            uint8 doff;
+            uint8 type;
+            uint16 channel;
+            Performative performative;
```
jfallows: Shouldn't this be `AmqpPerformative` from commented section above instead?

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -56,63 +105,140 @@
...
         this.router = requireNonNull(router);
         this.writeBuffer = requireNonNull(writeBuffer);
+        this.extBuffer = new UnsafeBuffer(new byte[writeBuffer.capacity()]);
+        this.frameBuffer = new UnsafeBuffer(new byte[writeBuffer.capacity()]);
+        this.boundedOctetsBuffer = new UnsafeBuffer(new byte[writeBuffer.capacity()]);
+        this.terminusListBuffer = new UnsafeBuffer(new byte[writeBuffer.capacity()]);
+        this.emptyListBuffer = new UnsafeBuffer(new byte[writeBuffer.capacity()]);
+        this.errorListBuffer = new UnsafeBuffer(new byte[writeBuffer.capacity()]);
+        this.stringBuffer = new UnsafeBuffer(new byte[writeBuffer.capacity()]);
+        this.messageBuffer = new UnsafeBuffer(new byte[writeBuffer.capacity()]);
```
jfallows: That's a lot of buffers. Are they all necessarily separate, or can we consolidate any of them?

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -144,18 +270,19 @@ private MessageConsumer newInitialStream(
...
 
         if (route != null)
         {
-            final AmqpServer connection = new AmqpServer(sender, routeId, initialId, replyId);
-            correlations.put(replyId, connection);
-            newStream = connection::onNetwork;
+            final long initialId = begin.streamId();
+            final long affinity = begin.affinity();
+            final long replyId = supplyReplyId.applyAsLong(initialId);
+            final long budgetId = supplyBudgetId.getAsLong();
```
jfallows: Assign `replyId` and `budgetId` inside `AmqpServer` instead.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -165,155 +292,847 @@ private MessageConsumer newReplyStream(
...
+        return correlations.remove(replyId);
     }
 
-    private RouteFW wrapRoute(
-        int msgTypeId,
+    private RouteFW resolveTarget(
+        long routeId,
+        long authorization,
+        String targetAddress,
+        AmqpCapabilities role)
```
jfallows: Rename `role` to `capabilities`.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -165,155 +292,847 @@ private MessageConsumer newReplyStream(
...
+        {
+            final RouteFW route = routeRO.wrap(b, o, o + l);
+            final OctetsFW ext = route.extension();
+            if (ext.sizeof() > 0)
+            {
+                final AmqpRouteExFW routeEx = ext.get(routeExRO::wrap);
+                final String targetAddressEx = routeEx.targetAddress().asString();
+                return targetAddressEx.equals(targetAddress) && routeEx.capabilities().get().equals(role);
+            }
+            return true;
```
jfallows:
```
final RouteFW route = routeRO.wrap(b, o, o + l);
final OctetsFW extension = route.extension();
final AmqpRouteExFW routeEx = extension.get(routeExRO::tryWrap);
return routeEx != null && Objects.equals(targetAddress, routeEx.targetAddress()) && capabilities == routeEx.capabilities().get();
```

Change `targetAddress` parameter type to `StringFW`, okay to pass `null` if no value available (better than passing `""`).

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -165,155 +292,847 @@ private MessageConsumer newReplyStream(
...
+            long authorization,
+            AmqpOpenFW amqpOpen)
         {
-            if (initialCredit > 0)
+            int offsetOpenFrame = amqpFrameHeaderRW.limit();
+
+            final String8FW containerIdRO = new String8FW.Builder()
+                .wrap(writeBuffer, 0, writeBuffer.capacity())
+                .set(containerId, UTF_8)
+                .build();
```
jfallows: Avoid creating `new` flyweights, hoist as fields on factory and reuse instead.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -326,323 +1145,1081 @@ private void onNetwork(
...
+        private void onDecodeOpen(
+            long traceId,
+            long authorization,
+            AmqpOpenFW open)
         {
-            // TODO
-            /*
-            * get a session from map with channel
-            * */
+            doEncodeOpen(traceId, authorization, open);
```
jfallows: Extract any information you need from `open` here, then call `doEncodeOpen` without passing `open`.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -326,323 +1145,1081 @@ private void onNetwork(
...
+                final MutableDirectBuffer buffer = bufferPool.buffer(encodeSlot);
+                final int limit = Math.min(encodeSlotOffset, encodeSlotMaxLimit);
+                final int maxLimit = encodeSlotOffset;
+
+                encodeNetwork(encodeSlotTraceId, authorization, budgetId, buffer, 0, limit, maxLimit);
+            }
+
+            final int slotCapacity = bufferPool.slotCapacity();
+            final int sharedReplyCredit = Math.min(slotCapacity, replyBudget - encodeSlotOffset - sharedReplyBudget);
+            AtomicInteger minRemoteIncomingWindow = new AtomicInteger();
```
jfallows: No need for `atomic`, consider `MutableInteger` instead, also hoist and share instance for reuse.
Alternatively, avoid using `forEach` and instead iterate with a `for` loop.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -326,323 +1145,1081 @@ private void onNetwork(
...
+            long traceId,
+            long authorization,
+            AmqpCloseFW close)
         {
-            // TODO
+            if (close.fieldCount() == 0)
+            {
+                sessions.values().forEach(s -> s.cleanup(traceId, authorization));
+                doEncodeClose(traceId, authorization, null);
+                doNetworkEnd(traceId, authorization);
```
jfallows: Needs conditional `doNetworkEnd` in case `encodeSlot` not fully encoded before closing network, e.g. not enough network window.
Also needs to set network `state` to mark reply as closing, then `encodeNetwork` can detect reply is closing and implicitly `doNetworkEnd` after `encodeSlot` is flushed.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -326,323 +1145,1081 @@ private void onNetwork(
...
+            {
+                byte[] nameInBytes = new byte[4];
+                buffer.getBytes(offset, nameInBytes, 0, 4);
+                return new String(nameInBytes);
+            });
+            return "AMQP".equals(name) &&
+                header.id() == 0 &&
+                header.major() == 1 &&
+                header.minor() == 0 &&
+                header.revision() == 0;
```
jfallows: This can be simplified as a `long` check for fixed `8 byte` value.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -326,323 +1145,1081 @@ private void onNetwork(
...
 
-        private void onAmqpEnd(
-            AmqpFrameFW end)
+        private void cleanupNetwork(
+            long traceId,
+            long authorization)
         {
-            // TODO
+            doNetworkReset(traceId, authorization);
+            doNetworkAbort(traceId, authorization);
```
jfallows: Make conditional versions of these in case the network is already half-closed.
```
doNetworkResetIfNecessary(traceId, authorization);
doNetworkAbortIfNecessary(traceId, authorization);
```
The condition is based on the network `state`.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -326,323 +1145,1081 @@ private void onNetwork(
...
+                encodeSlotTraceId = 0;
+            }
+        }
 
-            final AmqpFrameFW amqpFrame = amqpFrameRO.wrap(buffer, offset, offset + length);
+        private final class AmqpSession
+        {
+            private final Long2ObjectHashMap<AmqpServerStream> links;
+
+            private int channelId;
```
jfallows: `final` ?

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -326,323 +1145,1081 @@ private void onNetwork(
...
+                this.remoteOutgoingWindow = flowOutgoingWindow;
+                sharedReplyBudget = Math.min(remoteIncomingWindow * initialMaxFrameSize, replyBudget);
+
+                if (flow.hasHandle())
+                {
+                    links.get(flow.handle()).onDecodeFlow(traceId, authorization, deliveryCount, linkCredit);
+                }
+                else
+                {
+                    links.values().forEach(s -> s.onDecodeFlow(traceId, authorization, deliveryCount, linkCredit));
```
jfallows: The `links` don't need to process a `FLOW` frame just for the `session`, right?

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -326,323 +1145,1081 @@ private void onNetwork(
...
+                private long initialSlotTraceId;
+
+                private int state;
+                private Set<AmqpRole> roles;
+                private String name;
+                private long handle;
+
+                AmqpServerStream()
+                {
+                    this.roles = EnumSet.noneOf(AmqpRole.class);
```
jfallows: Use `AmqpCapabilities` capabilities mask, initialized to `0`, instead of `enum` `set`.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -326,323 +1145,1081 @@ private void onNetwork(
...
+                    case SENDER:
+                        route = resolveTarget(routeId, authorization, hasTargetAddress ?
+                            attach.target().targetList().address().asString() : "", AmqpCapabilities.SEND_ONLY);
+                        break;
+                    default:
+                        throw new IllegalStateException("Unexpected value: " + role);
+                    }
+
+                    if (route != null)
+                    {
```
jfallows: The part above should be done in `AmqpSession` before the stream is created, allowing it to be rejected directly by the `session` if no `route` is found.
If found, then construct `AmqpServerStream` passing `address` and `role` to constructor.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -326,323 +1145,1081 @@ private void onNetwork(
...
+                    assert !AmqpState.initialClosed(state);
+
+                    state = AmqpState.closedInitial(state);
+                    cleanupInitialSlotIfNecessary();
+
+                    if (AmqpState.closed(state))
+                    {
+                        sessions.clear();
+                        links.clear();
+                    }
```
jfallows, line 1953: Why are we clearing all `sessions` and `links` when an `application` stream is closed?

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -326,323 +1145,1081 @@ private void onNetwork(
...
+                    final int padding = window.padding();
+
+                    state = AmqpState.openedInitial(state);
+
+                    this.budgetId = budgetId;
+
+                    initialBudget += credit;
+                    initialPadding = padding;
+
+                    if (initialSlot != NO_SLOT)
```
jfallows: Is it actually possible to have a used `initialSlot` in `amqp`?
I think it may not be possible due to strictly enforced flow control. If not, then please remove it as having separate slots per application stream is a bit heavy.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -165,155 +292,847 @@ private MessageConsumer newReplyStream(
...
-            /*
-            final AmqpOpenFrameFW open = amqpOpenFrameRW.wrap(encodeBuffer, 0, encodeBuffer.capacity())
-                .length(frame.length())
-                .dof(frame.dof())
-                .type(frame.type())
-                .channel(frame.channel())
-                .performative(frame.performative())
-                .payload(frame.payload)
+            final AmqpDataExFW dataEx = extension.get(amqpDataExRO::tryWrap);
+            byte[] deliveryTagBytes = new byte[1];
```
jfallows: Avoid call to `new` especially on the data path as it creates garbage per message.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -165,155 +292,847 @@ private MessageConsumer newReplyStream(
...
+
+            int payloadSize = payload.sizeof();
+
+            final int frameHeaderSize = frameHeader.sizeof();
+            final int transferOffset = frameHeader.limit();
+            AmqpTransferFW.Builder transferBuilder =  amqpTransferRW.wrap(frameBuffer, transferOffset, frameBuffer.capacity());
+            AmqpTransferFW transfer = payloadIndex > 0 ? transferBuilder.handle(handle).build() :
+                transferBuilder.handle(handle)
+                    .deliveryId(dataEx.deliveryId())
+                    .deliveryTag(deliveryTag)
```
jfallows: Seems like you want a `.deliveryTag(OctetsFW)` setter as well, even though `deliveryTag` is a `BoundedOctetsFW`.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -165,155 +292,847 @@ private MessageConsumer newReplyStream(
...
+
+            payloadRW.put(payload.buffer(), payload.offset() + payloadIndex, payloadSize);
+            OctetsFW transferWithFrameHeader = payloadRW.build();
+            doNetworkData(traceId, authorization, 0L, transferWithFrameHeader);
+
+            if (hasMore)
+            {
+                doEncodeTransfer(traceId, authorization, channel, handle, flags, extension, payload,
+                    payloadIndex + payloadSize);
+            }
```
jfallows, line 998: This should be done as a loop until end of payload rather than as nested method calls.

See https://github.com/reaktivity/nukleus-http.java/blob/develop/src/main/java/org/reaktivity/nukleus/http2/internal/stream/Http2ServerFactory.java#L2392

jfallows (changes requested):

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
izzymoriguchi: When a size of `route.extension()` is 0 (in which `routeEx` becomes null), do we want to return `false`?

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
izzymoriguchi: > set network `state` to mark reply as closing

I'm not really sure where I should "set `state` to mark reply as closing". Could you elaborate a bit more?

jfallows (approved):

PR #11, opened 2020-04-27
Implement transfer with annotations and properties

izzymoriguchi:


src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -980,24 +1026,129 @@ private void doEncodeTransfer(
...
+        {
+            AmqpPropertiesFW properties = null;
+            if (propertiesArray.fieldCount() > 0)
+            {
+                amqpPropertiesRW.wrap(extraBuffer, bufferOffset, extraBuffer.capacity());
+                propertiesArray.forEach(field ->
+                {
+                    switch (field.kind())
+                    {
+                    case 0:
```
jfallows: Use constants from generated flyweight for `kind` values to make the code more readable.

jfallows (changes requested):

jfallows (approved):

PR #12, opened 2020-05-08
Implement application properties for data transfer

izzymoriguchi:


src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -1098,73 +1111,102 @@ private AmqpFrameHeaderFW setTransferFrame(
...
         }
 
         private AmqpPropertiesFW properties(
-            ArrayFW<AmqpMessagePropertyFW> propertiesArray,
+            AmqpDataExFW dataEx,
             int bufferOffset)
         {
             AmqpPropertiesFW properties = null;
-            if (propertiesArray.fieldCount() > 0)
+            if (dataEx.properties().fieldCount() > 0)
```
jfallows: Please hoist out `dataEx.properties()` perhaps called `propertiesEx`.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -1098,73 +1111,102 @@ private AmqpFrameHeaderFW setTransferFrame(
...
+                applicationProperties.forEach(pair ->
+                {
+                    AmqpStringFW key = amqpStringRW.wrap(valueBuffer, valueOffset.value, valueBuffer.capacity())
+                        .set(pair.key()).build();
+                    valueOffset.value += key.limit();
+                    AmqpStringFW value = amqpValueRW.wrap(valueBuffer, valueOffset.value, valueBuffer.capacity())
+                        .set(pair.value()).build();
+                    valueOffset.value += value.limit();
+                    amqpMapRW.entry(k -> k.setAsAmqpString(key), v -> v.setAsAmqpString(value));
+                });
```
jfallows: Can this be moved to a method and then pass the method reference in place of the lambda?

jfallows (changes requested):

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
izzymoriguchi: I kept using `dataEx.properties()` since there were name conflicts from `amqp.idl` and `protocol.idl` for `AmqpPropertiesFW`. I thought of using `AmqpTypeUtil` like other name conflicts but wasn't sure the best way to simplify this case.

jfallows: Can we change `protocol.idl` to use `AmqpMessageProperties` and `AmqpApplicationProperties`, leaving `AmqpProperties` for `amqp.idl`?

jfallows (approved):

PR #13, opened 2020-05-21
Update flow control and implement publish

izzymoriguchi:


src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -1702,7 +1728,11 @@ private void onDecodeOpen(
             long authorization,
             AmqpOpenFW open)
         {
-            doEncodeOpen(traceId, authorization, open.hasMaxFrameSize(), (int) open.maxFrameSize());
+            if (open.hasMaxFrameSize())
```
jfallows: If open max frame size has a default, does `open.maxFrameSize()` return that default when not explicitly present in the `open` frame?

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -2297,9 +2326,13 @@ private void onApplicationData(
                     remoteIncomingWindow--;
                     nextOutgoingId++;
                     outgoingWindow--;
-                    deliveryCount++;
-                    linkCredit--;
-                    doEncodeTransfer(traceId, authorization, channelId, handle, flags, extension, data.payload());
+                    if ((flags & 1) == 1)
```
jfallows: Define constants for `FLAG_INIT` `2`, `FLAG_FIN` `1` as necessary to avoid magic numbers in the code.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -2297,9 +2326,13 @@ private void onApplicationData(
...
+                    {
+                        deliveryCount++;
+                        linkCredit--;
+                    }
+                    doEncodeTransfer(traceId, authorization, channelId, remoteIncomingWindow, handle, flags, extension,
+                        data.payload());
```
jfallows, line 2857: Should we be flushing application window here after encode transfer?

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -2338,8 +2371,12 @@ private void flushReplyWindow(
                 {
                     if (isReplyOpen())
                     {
+                        final int slotCapacity = bufferPool.slotCapacity();
+                        final int maxNumberOfFrames = slotCapacity / initialMaxFrameSize +
+                            ((slotCapacity % initialMaxFrameSize == 0) ? 0 : 1);
+                        final int padding = 20 * maxNumberOfFrames + 205;
```
jfallows: Please define constants for `20` and `205` to avoid magic numbers in the code, also effectively describing the algorithm more clearly in code.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -1208,11 +1231,11 @@ private void putProperty(
...
             AmqpStringFW key = amqpStringRW.wrap(valueBuffer, valueOffset.value, valueBuffer.capacity())
                 .set(item.key()).build();
-            valueOffset.value += key.limit();
+            valueOffset.value += key.sizeof();
             AmqpStringFW value = amqpValueRW.wrap(valueBuffer, valueOffset.value, valueBuffer.capacity())
                 .set(item.value()).build();
-            valueOffset.value += value.limit();
-            amqpMapRW.entry(k -> k.setAsAmqpString(key), v -> v.setAsAmqpString(value));
+            valueOffset.value += value.sizeof();
+            applicationPropertyRW.entry(k -> k.setAsAmqpString(key), v -> v.setAsAmqpString(value));
```
jfallows: This style of coding is relying on side effects to update the `applicationPropertyRW` global variable across method invocations and then use the collected result afterwards.

While this is a good technique to take advantage of methods like `applicationProperties.forEach` by passing a consumer lambda, the scope is a bit broad as the variables are global.

Suggest creating inner classes for these cases, such as `AmqpApplicationProperties`, that manages the iteration, and has a local builder RW instance that is initially wrapped, then used for side effects before returning the collected result, all on stack for the same method invocation on `AmqpApplicationProperties`.

Suggest following that same pattern for message properties and annotations as well.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -1185,20 +1208,20 @@ private void putAnnotation(
...
                 AmqpSymbolFW name = amqpSymbolRW.wrap(valueBuffer, valueOffset.value, valueBuffer.capacity())
                     .set(item.key().name()).build();
-                valueOffset.value += name.limit();
+                valueOffset.value += name.sizeof();
                 AmqpByteFW value2 = amqpByteRW.wrap(valueBuffer, valueOffset.value, valueBuffer.capacity())
                     .set(item.value().bytes().value().getByte(0)).build();
-                valueOffset.value += value2.limit();
-                amqpMapRW.entry(k -> k.setAsAmqpSymbol(name), v -> v.setAsAmqpByte(value2));
+                valueOffset.value += value2.sizeof();
+                annotationRW.entry(k -> k.setAsAmqpSymbol(name), v -> v.setAsAmqpByte(value2));
```
jfallows: Let's use the corresponding generated `KIND` constants in the case statements instead of `1` and `2`.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -1615,7 +1641,7 @@ private void doNetworkWindow(
             state = AmqpState.openInitial(state);
 
             initialBudget += credit;
-            doWindow(network, routeId, initialId, traceId, authorization, budgetId, credit, padding);
+            doWindow(network, routeId, initialId, traceId, authorization, budgetId, credit, padding, 0);
```
jfallows, line 1870: Define a method called `doNetworkWindow` to make this part of the code more readable.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -2338,8 +2371,12 @@ private void flushReplyWindow(
                 {
                     if (isReplyOpen())
                     {
+                        final int slotCapacity = bufferPool.slotCapacity();
+                        final int maxNumberOfFrames = slotCapacity / initialMaxFrameSize +
+                            ((slotCapacity % initialMaxFrameSize == 0) ? 0 : 1);
```
jfallows: Try to avoid `%` operator as it is quite heavy.
Does `final int maxNumberOfFrames = (slotCapacity + initialMaxFrameSize - 1) / initialMaxFrameSize;` work?

jfallows (changes requested):
After you've applied this feedback, let's look at `doEncodeTransfer` together in a second pass to see if we can simplify the logic.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
izzymoriguchi: Yes, `open.maxFrameSize()` returns a dafault value when the value is not explicityly present.
`if (open.hasMaxFrameSize())` returns true only when the value is explicitly set, so I don't think we need this check, so I'll remove this `if` statement.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
izzymoriguchi: Yes, that makes sense to me.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
izzymoriguchi: This `doWindow(..)` call is made inside `doNetworkWindow` method. Are you suggesting to create another `doNetworkWindow` method that takes in different parameters?

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
jfallows: 👍

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
jfallows: sorry, that context was lost on me in the diff, it's all good - already done 👍

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -920,134 +925,132 @@ private void doEncodeAttach(
...
         {
             final AmqpDataExFW dataEx = extension.get(amqpDataExRO::tryWrap);
-            int deferred = dataEx != null ? dataEx.deferred() : 0;
-            int extraBufferOffset = 0;
-            int flag = dataEx != null ? dataEx.flags() : 0;
-            int bitmask = 1;
-            int settled = 0;
-            if ((flag & bitmask) == bitmask)
+            int numberOfTransfer = 0;
+            if (dataEx != null) // TODO: Means flags = INIT_ONLY or BOTH
```
jfallows: For `TODO` perhaps add `assert (flags & FLAG_INIT) != 0;` inside the `if` block.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -920,134 +925,132 @@ private void doEncodeAttach(
...
-                            .deliveryId(dataEx.deliveryId())
-                            .deliveryTag(deliveryTag)
-                            .messageFormat(dataEx.messageFormat())
-                            .settled(settled)
-                            .build();
-                final int transferFrameSize = transfer.sizeof() + Byte.BYTES;
-                payloadRW.wrap(writeBuffer, DataFW.FIELD_OFFSET_PAYLOAD, writeBuffer.capacity());
-                if (payloadIndex == 0)
+                final int deferred = dataEx.deferred();
+                final int transferFlag = dataEx.flags();
```
jfallows: `transferFlags` (plural)

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -434,6 +466,7 @@ private void doWindow(
             .budgetId(budgetId)
             .credit(credit)
             .padding(padding)
+            .minimum(min)
```
jfallows: let's rename `min` to `minimum` to match

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -586,7 +618,13 @@ private int decodeOpen(
 
         int progress = offset;
 
-        if (open != null)
+        if (limit - offset > defaultMaxFrameSize)
+        {
+            server.onDecodeError(traceId, authorization, CONNECTION_FRAMING_ERROR);
+            server.decoder = decodeFrameType;
+            progress = limit;
+        }
```
jfallows: Can we move this check into `decodeFrameType` instead?
That would let us check once when each frame arrives.

Seems like this should be comparing the actual frame size (not limit - offset, which might represent multiple frames) to `initialMaxFrameSize`.
Here the term `initialMaxFrameSize` refers to the `initial` direction (vs `reply` direction) so this can be defaulted to the spec-defined default for max frame size until we send the `OPEN` frame to the client, then we can overwrite `initialMaxFrameSize` to the configured value. This will ensure we are enforcing the max frame size properly for all frames, both initially and after changing from the default to the configured value.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -656,15 +714,254 @@ private int decodeFlow(
...
+
+            if (transfer.hasDeliveryId())
+            {
+                AmqpDescribedTypeFW describedType = amqpDescribedTypeRO.tryWrap(buffer, payloadOffset, limit);
+                while (describedType != null)
+                {
+                    if (describedType.get() == VALUE)
+                    {
+                        break;
+                    }
```
jfallows: Put `loop:` above the `while`, then remove this `if`, add a `case` for `VALUE` into the `switch` and change the `VALUE` `break;` to `break loop;` to exit the `while` loop for the `VALUE` case without extracting the `describedType` after the `switch`.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -656,15 +714,254 @@ private int decodeFlow(
...
+            int payloadOffset = transfer.limit();
+            final DirectBuffer transferBuffer = transfer.buffer();
+            Array32FW<AmqpAnnotationFW> annotations = null;
+            AmqpPropertiesFW properties = null;
+            Array32FW<AmqpApplicationPropertyFW> applicationProperties = null;
+            int annotationSize = 0;
+            int propertySize = 0;
+            int applicationPropertySize = 0;
+
+            if (transfer.hasDeliveryId())
```
jfallows: By spec, `deliveryId` can be present on all `TRANSFER` frames.

> The delivery-id MUST be supplied on the first transfer of a multi-transfer delivery. On continuation transfers the delivery-id MAY be omitted. It is an error if the delivery-id on a continuation transfer differs from the delivery-id on the first transfer of a delivery.

Would tracking the current `deliveryId` on the `sender` stream be helpful here instead? Storing as a `long` would allow for all `32 bits` plus leave room for `-1` as an "null" value to be used between messages on the same `link`.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -656,15 +714,254 @@ private int decodeFlow(
...
+            progress = limit;
+            server.decoder = decodeFrameType;
+        }
+        else if (transfer != null)
+        {
+            AmqpServer.AmqpSession.AmqpServerStream sender =
+                server.sessions.get(amqpFrameHeaderRO.channel()).links.get(transfer.handle());
+
+            int payloadOffset = transfer.limit();
+            final DirectBuffer transferBuffer = transfer.buffer();
```
jfallows: This can be removed and use `buffer` instead since `transferBuffer == buffer`.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -656,15 +714,254 @@ private int decodeFlow(
...
+            Array32FW<AmqpAnnotationFW> annotations = null;
+            AmqpPropertiesFW properties = null;
+            Array32FW<AmqpApplicationPropertyFW> applicationProperties = null;
+            int annotationSize = 0;
+            int propertySize = 0;
+            int applicationPropertySize = 0;
+
+            if (transfer.hasDeliveryId())
+            {
+                AmqpDescribedTypeFW describedType = amqpDescribedTypeRO.tryWrap(buffer, payloadOffset, limit);
```
jfallows: Rename to `sectionType` ?

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -656,15 +714,254 @@ private int decodeFlow(
...
+            server.onDecodeError(traceId, authorization, CONNECTION_FRAMING_ERROR);
+            progress = limit;
+            server.decoder = decodeFrameType;
+        }
+        else if (transfer != null)
+        {
+            AmqpServer.AmqpSession.AmqpServerStream sender =
+                server.sessions.get(amqpFrameHeaderRO.channel()).links.get(transfer.handle());
+
+            int payloadOffset = transfer.limit();
```
jfallows: Rename to `sectionOffset` ?

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -656,15 +714,254 @@ private int decodeFlow(
...
+                    if (describedType.get() == VALUE)
+                    {
+                        break;
+                    }
+                    switch (describedType.get())
+                    {
+                    case MESSAGE_ANNOTATIONS:
+                        AmqpMapFW<AmqpValueFW, AmqpValueFW> annotation =
+                            annotationRO.tryWrap(buffer, describedType.limit(), limit);
+                        if (annotation != null)
```
jfallows: Is it a protocol error at this point if `annotation` is `null` ?

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -656,15 +714,254 @@ private int decodeFlow(
...
+                        {
+                            payloadOffset = annotation.limit();
+                            annotations = decodeAnnotations(annotation);
+                            annotationSize = describedType.sizeof() + annotation.sizeof();
+                        }
+                        break;
+                    case PROPERTIES:
+                        AmqpMessagePropertiesFW property = amqpPropertiesRO.tryWrap(buffer, describedType.limit(), limit);
+                        if (property != null)
```
jfallows: Is it a protocol error at this point if `property` is `null` ?

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -656,15 +714,254 @@ private int decodeFlow(
...
+                        {
+                            payloadOffset = property.limit();
+                            properties = decodeProperties(property);
+                            propertySize = describedType.sizeof() + properties.sizeof();
+                        }
+                        break;
+                    case APPLICATION_PROPERTIES:
+                        AmqpMapFW<AmqpValueFW, AmqpValueFW> applicationProperty =
+                            applicationPropertyRO.tryWrap(buffer, describedType.limit(), limit);
+                        if (applicationProperty != null)
```
jfallows: Is it a protocol error at this point if `applicationProperty` is `null` ?

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -656,15 +714,254 @@ private int decodeFlow(
...
+                        {
+                            payloadOffset = applicationProperty.limit();
+                            applicationProperties = decodeApplicationProperties(applicationProperty);
+                            applicationPropertySize = describedType.sizeof() + applicationProperties.sizeof();
+                        }
+                        break;
+                    }
+                    describedType = amqpDescribedTypeRO.tryWrap(buffer, payloadOffset, limit);
+                }
+            }
```
jfallows, line 693: Actually, I think these sections actually have fixed order by spec, so we probably don't want a loop, just linear processing in spec order.

- Zero or one header.
- Zero or one delivery-annotations.
- Zero or one message-annotations.
- Zero or one properties.
- Zero or one application-properties.
- The body consists of one of the following three choices: one or more data sections, one or more amqp-sequence sections, or a single amqp-value section.
-  Zero or one footer. 

where a data section is amqp-binary.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -656,15 +714,254 @@ private int decodeFlow(
...
+    {
+        Array32FW.Builder<AmqpAnnotationFW.Builder, AmqpAnnotationFW> annotationBuilder =
+            annotationRW.wrap(frameBuffer, 0, frameBuffer.capacity());
+        annotation.forEach(kv -> vv ->
+        {
+            BoundedOctetsFW value = vv.getAsAmqpBinary().get();
+            switch (kv.kind())
+            {
+            case SYMBOL1:
+                String symbolKey = kv.getAsAmqpSymbol().get().asString();
```
jfallows: Can we keep the `symbolKey` as a `StringFW` to avoid the UTF8 decode and encode?

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -656,15 +714,254 @@ private int decodeFlow(
...
+        }
+        if (property.hasGroupSequence())
+        {
+            propertyBuilder.groupSequence((int) property.groupSequence());
+        }
+        if (property.hasReplyToGroupId())
+        {
+            propertyBuilder.replyToGroupId(property.replyToGroupId().asString());
+        }
+        return propertyBuilder.build();
```
jfallows, line 856: Can the `.asString()` calls be removed above?

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -656,15 +714,254 @@ private int decodeFlow(
...
+                            applicationProperties = decodeApplicationProperties(applicationProperty);
+                            applicationPropertySize = describedType.sizeof() + applicationProperties.sizeof();
+                        }
+                        break;
+                    }
+                    describedType = amqpDescribedTypeRO.tryWrap(buffer, payloadOffset, limit);
+                }
+            }
+
+            int reserved;
```
jfallows: Let's review `reserved` once the above feedback is in place. It should also include any necessary `padding` for the application initial stream.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -925,102 +1285,122 @@ private void doEncodeTransfer(
...
-                    amqpSection(annotations, properties, applicationProperties, sectionOffset);
-                    payloadRW.put(valueHeader.buffer(), valueHeader.offset(), valueHeader.sizeof());
-                }
-                else
+                final int deferred = dataEx.deferred();
+                final int transferFlag = dataEx.flags();
+                final boolean settled = (transferFlag & FLAG_SETTLED) != 0;
+                final boolean resume = (transferFlag & FLAG_RESUME) != 0;
+                final boolean aborted = (transferFlag & FLAG_ABORTED) != 0;
+                final boolean batchable = (transferFlag & FLAG_BATCHABLE) != 0;
```
jfallows: Suggest hoisting these 4 lines off into an `AmqpTransferFlags` helper class with static methods for `isSettled(int transferFlags)`, etc that can be called when needed below.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -925,102 +1285,122 @@ private void doEncodeTransfer(
...
-            int extraBufferOffset = 0;
-            final BoundedOctetsFW deliveryTag = amqpBinaryRW.wrap(extraBuffer, extraBufferOffset, extraBuffer.capacity())
-                .set(dataEx.deliveryTag().bytes().value(), 0, dataEx.deliveryTag().length())
-                .build()
-                .get();
-            int flag = dataEx.flags();
-            int bitmask = 1;
-            int settled = 0;
-            if ((flag & bitmask) == bitmask)
+            int numberOfTransfer = 0;
```
jfallows: Rename to `transferFrames`.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -1856,6 +2219,87 @@ private void cleanupEncodeSlotIfNecessary()
             }
         }
 
+        private final class AmqpAnnotations
+        {
+            private final AmqpMapFW.Builder<AmqpValueFW, AmqpValueFW, AmqpValueFW.Builder, AmqpValueFW.Builder> annotationRW =
```
jfallows: Rename to `annotationsRW` (plural)

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -1856,6 +2219,87 @@ private void cleanupEncodeSlotIfNecessary()
...
+                new AmqpMapFW.Builder<>(new AmqpValueFW(), new AmqpValueFW(), new AmqpValueFW.Builder(),
+                    new AmqpValueFW.Builder());
+
+            private AmqpMapFW<AmqpValueFW, AmqpValueFW> annotations(
+                AmqpDataExFW dataEx,
```
jfallows: Change parameter to `ArrayFW<AmqpAnnotationFW> annotations` then pass `dataEx.annotations()` at call site instead of `dataEx`.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -1856,6 +2219,87 @@ private void cleanupEncodeSlotIfNecessary()
...
+                return null;
+            }
+
+            private void annotation(
+                AmqpAnnotationFW item)
+            {
+                switch (item.key().kind())
+                {
+                case KIND_ID:
+                    AmqpULongFW id = amqpULongRW.wrap(valueBuffer, valueOffset.value, valueBuffer.capacity())
```
jfallows: Remove `valueOffset` mutable integer and use a `valueOffset` `int` field here in `AmqpAnnotations` instead, then merge the `AmqpAnnotations` and `AmqpApplicationProperties ` helper classes together as needed to avoid unnecessary global variable references and side effects.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -1945,23 +2389,18 @@ private void onDecodeAttach(
...
-                                address = attach.source().sourceList().address().asString();
-                            }
-                            break;
-                        case SENDER:
-                            if (hasTargetAddress)
-                            {
-                                address = attach.target().targetList().address().asString();
-                            }
-                            break;
+                            addressFrom = attach.source().sourceList().address().asString();
```
jfallows: Try to avoid repeated execution of these nested method calls.

For example, perhaps hoist `attach.source().sourceList()` into a local variable `source` that might be `null` if not `attach.hasSource()`, then `if (source != null)` replaces `if (hasSourceAddress)` above and `addressFrom = source.address().asString();` replaces `addressFrom = attach.source().sourceList().address().asString();` too, which seems to improve readability.

Also, can `addressFrom` and `addressTo` be kept as `StringFW` instead of `String` to avoid UTF8 overhead?

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -1945,23 +2389,18 @@ private void onDecodeAttach(
...
                         }
-                        AmqpServerStream link = new AmqpServerStream(address, role, route);
+                        if (hasTargetAddress)
+                        {
+                            addressTo = attach.target().targetList().address().asString();
```
jfallows: Similar feedback as source address above for target address here.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -1997,23 +2435,46 @@ private void onDecodeFlow(
...
+                final BoundedOctetsFW deliveryTag = transfer.hasDeliveryTag() ? transfer.deliveryTag() : null;
+                final long messageFormat = transfer.hasMessageFormat() ? transfer.messageFormat() : 0;
+                final int settled = transfer.hasSettled() ? transfer.settled() : 0;
+                final int resume = transfer.hasResume() ? transfer.resume() : 0;
+                final int aborted = transfer.hasAborted() ? transfer.aborted() : 0;
+                final int batchable = transfer.hasBatchable() ? transfer.batchable() : 0;
+                int flags = settled == 1 ? FLAG_SETTLED : 0;
+                flags = resume == 1 ? flags | FLAG_RESUME : flags;
+                flags = aborted == 1 ? flags | FLAG_ABORTED : flags;
+                flags = batchable == 1 ? flags | FLAG_BATCHABLE : flags;
```
jfallows: Use `AmqpTransferFlags` helper class static methods such as `int settled(int transferFlags)`, etc.

The conditional logic can also seem to be simplified a little bit.
```
int flags = 0;
if (transfer.settled() == 1)
{
    flags = AmqpTransferFlags.settled(flags);
}
if (transfer.resume() == 1)
{
    flags = AmqpTransferFlags.resume(flags);
}
if (transfer.aborted() == 1)
{
    flags = AmqpTransferFlags.aborted(flags);
}
if (transfer.batchable() == 1)
{
    flags = AmqpTransferFlags.batchable(flags);
}
```

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -1997,23 +2435,46 @@ private void onDecodeFlow(
...
+                AmqpServerStream attachedLink = links.get(transfer.handle());
+                if (!transfer.hasMore())
```
jfallows: What if `more` is present on the `transfer` but explicitly set to `false`?

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -1997,23 +2435,46 @@ private void onDecodeFlow(
...
+                final long messageFormat = transfer.hasMessageFormat() ? transfer.messageFormat() : 0;
+                final int settled = transfer.hasSettled() ? transfer.settled() : 0;
+                final int resume = transfer.hasResume() ? transfer.resume() : 0;
+                final int aborted = transfer.hasAborted() ? transfer.aborted() : 0;
+                final int batchable = transfer.hasBatchable() ? transfer.batchable() : 0;
+                int flags = settled == 1 ? FLAG_SETTLED : 0;
+                flags = resume == 1 ? flags | FLAG_RESUME : flags;
+                flags = aborted == 1 ? flags | FLAG_ABORTED : flags;
+                flags = batchable == 1 ? flags | FLAG_BATCHABLE : flags;
+                AmqpServerStream attachedLink = links.get(transfer.handle());
```
jfallows: Rename to `link`?

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -2269,10 +2822,34 @@ private void onApplicationBegin(
...
+                            break;
+                        case RECEIVER:
+                            otherAddress = addressFrom;
+                            break;
+                        }
 
                         deliveryCount = initialDeliveryCount;
                         doEncodeAttach(traceId, authorization, name, channelId, handle, amqpRole, amqpSenderSettleMode,
-                            amqpReceiverSettleMode, address);
+                            amqpReceiverSettleMode, address, otherAddress);
```
jfallows, line 2807: What if `amqpBeginEx` is `null` ?

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -2269,10 +2822,34 @@ private void onApplicationBegin(
...
+
+                        if (amqpRole == RECEIVER)
+                        {
+                            this.linkCredit = (int) (Math.min(bufferPool.slotCapacity(), initialBudget) /
+                                Math.min(bufferPool.slotCapacity(), defaultMaxFrameSize));
+                            minimum.value = Integer.MAX_VALUE;
+                            links.values().forEach(l -> minimum.value = l.linkCredit == 0 ? minimum.value :
+                                Math.min(l.linkCredit, minimum.value));
+                            incomingWindow = minimum.value == Integer.MAX_VALUE ? 0 : minimum.value;
```
jfallows, line 2816: I think this part is going to change in your upcoming PR, so will not comment any further for now.

jfallows (changes requested):

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
izzymoriguchi: If we want to remove those calls we need to cast `property.replyToGroupId()` since the type is `StringFW` and the builder class's accessor takes `String8FW`

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
izzymoriguchi: To set each annotation item and application properties item, I'm currently using the same buffer (valueBuffer), and I'm concerned that if we don't use the global variable as an offset, it might be an issue when setting both annotation and application properties items. I think I had that issue before, so that is why I initially made the offset global, but now I don't have spec scripts that set both annotations and application properties at once so I don't have an error.

jfallows: It will still effectively be global since the merged `AmqpAnnotations` and `AmqpApplicationProperties` class instance will still be global, right?

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
jfallows: I see, would also prefer to avoid the cast if we can. 😉 

Would it make sense to adjust the builder to take `StringFW` instead of `String8FW` and then trigger an `assert` if the length is too big?

That would seem to still work for `String8FW` but also permit generic `StringFW` and even sufficiently short `String16FW` and `String32FW` to work as well.

We can go with the cast for now, to get the UTF8 benefit, and take the builder enhancement as a secondary step.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
izzymoriguchi: I believe it becomes `amqp:decode-error` if `annotation` is `null`?

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
izzymoriguchi: Oh I see!

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
izzymoriguchi: Yes it makes sense for the builder to take `StringFW` I think.

I took a note for future enhancement.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
jfallows: Agreed.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
jfallows: 👍

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
izzymoriguchi: To achieve that I think I need to update `union` builder to have a mutator that takes in `StringFW`. Should I do it?

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
izzymoriguchi: Would it be `amqp:decode-error` ?

jfallows: i think it just means we go with defaults as required by the protocol, since there was no issue with the decode that produced the initial amqp begin.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
jfallows: Yes, but not needed for this PR.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -1690,19 +2021,20 @@ private void onDecodeFrameHeader(
...
         }
 
         private void onDecodeOpen(
             long traceId,
             long authorization,
             AmqpOpenFW open)
         {
-            doEncodeOpen(traceId, authorization, open.hasMaxFrameSize(), (int) open.maxFrameSize());
+            this.encodeMaxFrameSize = Math.min(replySharedBudget, open.maxFrameSize());
+            doEncodeOpen(traceId, authorization, open.hasMaxFrameSize());
```
jfallows: Can we remove the `hasMaxFrameSize` parameter?
Our decision to send max frame size seems like it should not depend on whether or not the client sent us a max frame size, right?

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -1690,19 +2021,20 @@ private void onDecodeFrameHeader(
...
-                    .size(0)
-                    .doff(frameHeader.doff())
-                    .type(frameHeader.type())
-                    .channel(frameHeader.channel())
-                    .performative(mutator);
+                .size(0)
+                .doff(frameHeader.doff())
+                .type(frameHeader.type())
+                .channel(frameHeader.channel())
+                .performative(mutator);
```
jfallows: Seems like this is either not needed or else we are relying on a side effect of modifying global state in `amqpFrameHeaderRW`.

Suggest we try to remove the need for this method.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -498,54 +537,62 @@ private int decodeFrameType(
...
+                    frame = performative.end();
+                    server.onDecodeFrameHeader(frameHeader, b -> b.end(performative.end()));
+                    break;
+                case CLOSE:
+                    frame = performative.close();
+                    server.onDecodeFrameHeader(frameHeader, b -> b.close(performative.close()));
+                    break;
+                default:
+                    throw new IllegalStateException("Unexpected value: " + frameType);
+                }
```
jfallows: Suggest we remove this switch statement, extract whatever information we need from the frameHeader as decode member fields, and use them in `decodeOpen`, etc.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -588,9 +635,18 @@ private int decodeOpen(
 
         if (open != null)
```
jfallows: If the `else` clause is expected to be impossible due to earlier decode frame header logic, then we should `assert open != null` instead, same for other performatives.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -198,34 +231,36 @@
...
     private final AmqpServerDecoder decodeFlow = this::decodeFlow;
+    private final AmqpServerDecoder decodeTransfer = this::decodeTransfer;
     private final AmqpServerDecoder decodeClose = this::decodeClose;
     private final AmqpServerDecoder decodeIgnoreAll = this::decodeIgnoreAll;
     private final AmqpServerDecoder decodeUnknownType = this::decodeUnknownType;
 
     private final int outgoingWindow;
     private final String containerId;
-    private final long defaultMaxFrameSize;
+    private final long decodeMaxFrameSize;
```
jfallows: this field should be called `maxFrameSize` to match the configuration setting
`decodeMaxFrameSize` should be a member field on the `AmqpServer`, initialized to `MIN_MAX_FRAME_SIZE`.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -588,9 +635,18 @@ private int decodeOpen(
...
         {
-            server.onDecodeOpen(traceId, authorization, open);
-            server.decoder = decodeFrameType;
-            progress = open.limit();
+            if (open.sizeof() + FRAME_HEADER_SIZE > MIN_MAX_FRAME_SIZE)
+            {
+                server.onDecodeError(traceId, authorization, CONNECTION_FRAMING_ERROR);
+                server.decoder = decodeFrameType;
+                progress = limit;
+            }
```
jfallows: This check should be removed, as it will be covered by initializing the `AmqpServer` member field `decodeMaxFrameSize` to `MIN_MAX_FRAME_SIZE`.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -1690,19 +2021,20 @@ private void onDecodeFrameHeader(
...
+                .performative(mutator);
         }
 
         private void onDecodeOpen(
             long traceId,
             long authorization,
             AmqpOpenFW open)
         {
-            doEncodeOpen(traceId, authorization, open.hasMaxFrameSize(), (int) open.maxFrameSize());
+            this.encodeMaxFrameSize = Math.min(replySharedBudget, open.maxFrameSize());
```
jfallows, line 1971: Shouldn't `replySharedBudget` be  replaced by `bufferPool.slotCapacity()` here?

Also, this is where `AmqpServer` member field `decodeMaxFrameSize` should be set to `maxFrameSize` taken from config.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -801,9 +1111,8 @@ private void doEncodeOpen(
 
             if (hasMaxFrameSize)
             {
-                amqpOpenRW.maxFrameSize(defaultMaxFrameSize);
+                amqpOpenRW.maxFrameSize(decodeMaxFrameSize);
```
jfallows: The condition on whether or not to set `amqpOpenRW.maxFrameSize(...)` should depend only on the value. If it is not matching the default according to spec, then it should be included, otherwise omitted for brevity.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -1473,14 +1777,17 @@ private void onNetworkWindow(
...
+                (int) Math.min(minimum.value * encodeMaxFrameSize, replyBudget) : replyBudget;
+            final int replySharedCredit = replySharedBudgetMax - Math.max(replySharedBudget, 0) - Math.max(encodeSlotOffset, 0);
+
+            if (replySharedCredit > 0)
             {
-                final long replySharedBudgetPrevious = creditor.credit(traceId, replyBudgetIndex, sharedReplyCredit);
-                sharedReplyBudget = Math.min(minimum.value * initialMaxFrameSize, replyBudget);
+                final long replySharedBudgetPrevious = creditor.credit(traceId, replyBudgetIndex, replySharedCredit);
 
+                replySharedBudget += replySharedCredit;
```
jfallows: Take a look at recent changes to `http2` reply shared budget and make adjustments to above accordingly.

jfallows (changes requested):

jfallows (dismissed):

jfallows (approved):

PR #20, opened 2020-08-11
Implement bodyKind for amqp message body types

### LINKED ISSUE #21: Support message body fragmentation for data, sequence and value types
Allow message sections to span `TRANSFER` frames as needed during encode and decode.

See section [3.2 Message Format](http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-messaging-v1.0-os.html#section-message-format)

izzymoriguchi:
Fix #21

### DIFF
--- a/src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
+++ b/src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
@@ -31,12 +31,14 @@ import static org.reaktivity.nukleus.amqp.internal.types.codec.AmqpDescribedType
 import static org.reaktivity.nukleus.amqp.internal.types.codec.AmqpDescribedType.ATTACH;
 import static org.reaktivity.nukleus.amqp.internal.types.codec.AmqpDescribedType.BEGIN;
 import static org.reaktivity.nukleus.amqp.internal.types.codec.AmqpDescribedType.CLOSE;
+import static org.reaktivity.nukleus.amqp.internal.types.codec.AmqpDescribedType.DATA;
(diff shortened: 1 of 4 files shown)


src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -2531,6 +2627,32 @@ private void onDecodeTransfer(
...
+                            case STRING4:
+                                bodyKind = AmqpBodyKind.VALUE_STRING;
+                                break;
+                            case BINARY1:
+                            case BINARY4:
+                                bodyKind = AmqpBodyKind.VALUE_BINARY;
+                            }
+                            break;
+                        }
+                    }
```
jfallows: Recommend hoisting this to a `static` helper method, passing `AmqpSectionFW` returning `AmqpBodyKind`.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -802,6 +807,94 @@ private int decodeTransfer(
...
+        {
+            totalPayloadSize = amqpSection.data().sizeof();
+        }
+        return totalPayloadSize;
+    }
+
+    private long sectionHeaderSize(
+        AmqpSectionFW amqpSection)
+    {
+        long headerSize = amqpSection.limit();
```
jfallows: size and limit are different concepts, normally size would be limit - offset
suggest reworking this as `headerOffset` and `headerLimit`, then returning `headerLimit - headerOffset` at the end to get header size.

jfallows (changes requested):

jfallows (dismissed):

jfallows (approved):

PR #25, opened 2020-08-28
Update implementation to change deliveryId to session-scoped

### LINKED ISSUE #22: Include delivery id sequence number per message
Receiver expects each message to have an increasing `delivery-id` sequence number.

izzymoriguchi:
Fix #22

### DIFF
--- a/src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
+++ b/src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
@@ -156,7 +156,6 @@ public final class AmqpServerFactory implements StreamFactory
     private static final int FLAG_INIT_AND_FIN = 3;
     private static final int FRAME_HEADER_SIZE = 11;
     private static final int MIN_MAX_FRAME_SIZE = 512;
-    private static final int DESCRIBED_TYPE_SIZE = 3;
     private static final int TRANSFER_HEADER_SIZE = 20;
     private static final int PAYLOAD_HEADER_SIZE = 205;
     private static final int NO_DELIVERY_ID = -1;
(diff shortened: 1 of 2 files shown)


src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -795,17 +794,6 @@ private int decodeTransfer(
...
-            {
-                assert deliveryId != NO_DELIVERY_ID; // TODO: error
-                assert deliveryId == sender.deliveryId + 1; // TODO: error
-                sender.deliveryId = deliveryId;
-            }
-
-            if (deliveryId != NO_DELIVERY_ID && deliveryId != sender.deliveryId) // TODO: error
-            {
-                break decode;
-            }
```
jfallows, line 808: Looks like we lost the check for invalid inbound `deliveryId`, this needs to be restored, right?
Please make sure to separate `remoteDeliveryId` coming from the client towards us from `deliveryId` generated by us, heading towards the client.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -2192,19 +2180,8 @@ private void onDecodeTransfer(
...
-                    {
-                        link.onDecodeError(traceId, authorization, LINK_TRANSFER_LIMIT_EXCEEDED);
-                    }
-                    else
-                    {
-                        link.onDecodeTransfer(traceId, authorization, reserved, deliveryId, deliveryTag, messageFormat, settled,
-                            resume, aborted, batchable, more, buffer, offset, limit);
-                    }
+                    link.onDecodeTransfer(traceId, authorization, reserved, deliveryTag, messageFormat, settled,
+                        resume, aborted, batchable, more, buffer, offset, limit);
```
jfallows, line 2195: Do we need to keep these checks to validate inbound `deliveryId` or is it fully handled inside the link instead now?

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -2674,11 +2656,11 @@ private void onApplicationDataInit(
                     final AmqpDataExFW dataEx = extension.get(amqpDataExRO::tryWrap);
                     assert dataEx != null;
 
+                    deliveryId++;
```
jfallows: Suggest moving this just above the call to `onApplicationDataInit` in `onApplicationData`.

jfallows (changes requested):

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
izzymoriguchi: I believe I handled it inside the link's `onDecodeTransfer`

jfallows (dismissed):

jfallows (approved):

PR #36, opened 2020-09-04
Fix sending null messages

izzymoriguchi:
Requires https://github.com/reaktivity/nukleus-amqp.spec/pull/25

### DIFF
--- a/src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
+++ b/src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
@@ -188,7 +188,7 @@ public final class AmqpServerFactory implements StreamFactory
     private final OctetsFW.Builder payloadRW = new OctetsFW.Builder();
     private final OctetsFW.Builder messageFragmentRW = new OctetsFW.Builder();
 
-    private final OctetsFW octetsRO = new OctetsFW();
+    private final OctetsFW payloadRO = new OctetsFW();
 
     private final AmqpProtocolHeaderFW amqpProtocolHeaderRO = new AmqpProtocolHeaderFW();
     private final AmqpFrameHeaderFW amqpFrameHeaderRO = new AmqpFrameHeaderFW();
@@ -235,29 +235,37 @@ public final class AmqpServerFactory implements StreamFactory
         new Array32FW.Builder<>(new AmqpApplicationPropertyFW.Builder(), new AmqpApplicationPropertyFW());
 
     private final AmqpDescribedTypeFW applicationPropertiesSectionType = new AmqpDescribedTypeFW.Builder()
-            .wrap(new UnsafeBuffer(new byte[3]), 0, 3)
-            .set(APPLICATION_PROPERTIES)
-            .build();
+        .wrap(new UnsafeBuffer(new byte[3]), 0, 3)
+        .set(APPLICATION_PROPERTIES)
+        .build();
 
     private final AmqpDescribedTypeFW messagePropertiesSectionType = new AmqpDescribedTypeFW.Builder()
-            .wrap(new UnsafeBuffer(new byte[3]), 0, 3)
-            .set(PROPERTIES)
-            .build();
+        .wrap(new UnsafeBuffer(new byte[3]), 0, 3)
+        .set(PROPERTIES)
+        .build();
 
     private final AmqpDescribedTypeFW messageAnnotationsSectionType = new AmqpDescribedTypeFW.Builder()
-            .wrap(new UnsafeBuffer(new byte[3]), 0, 3)
-            .set(MESSAGE_ANNOTATIONS)
-            .build();
+        .wrap(new UnsafeBuffer(new byte[3]), 0, 3)
+        .set(MESSAGE_ANNOTATIONS)
+        .build();
 
     private final AmqpDescribedTypeFW dataSectionType = new AmqpDescribedTypeFW.Builder()
-            .wrap(new UnsafeBuffer(new byte[3]), 0, 3)
-            .set(DATA)
-            .build();
-
-    private final OctetsFW nullConstructor = new OctetsFW.Builder()
-        .wrap(new UnsafeBuffer(new byte[1]), 0, 1)
-        .set(new byte[] {0x40})
+        .wrap(new UnsafeBuffer(new byte[3]), 0, 3)
+        .set(DATA)
         .build();
+
+    private final AmqpDescribedTypeFW sequenceSectionType = new AmqpDescribedTypeFW.Builder()
+        .wrap(new UnsafeBuffer(new byte[3]), 0, 3)
+        .set(SEQUENCE)
+        .build();
+
+    private final AmqpDescribedTypeFW valueSectionType = new AmqpDescribedTypeFW.Builder()
+        .wrap(new UnsafeBuffer(new byte[3]), 0, 3)
+        .set(VALUE)
+        .build();
+
+    private final OctetsFW nullConstructor = new OctetsFW()
+        .wrap(new UnsafeBuffer(new byte[] {0x40}), 0, 1);
 
     private final AmqpMessageEncoder amqpMessageHelper = new AmqpMessageEncoder();
     private final AmqpMessageDecoder amqpMessageDecodeHelper = new AmqpMessageDecoder();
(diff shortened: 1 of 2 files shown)


src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -2426,9 +2437,9 @@ private void doApplicationData(
                 {
                     assert AmqpState.initialOpening(state);
 
-                    final DirectBuffer buffer = payload.buffer();
-                    final int offset = payload.offset();
-                    final int limit = payload.limit();
+                    final DirectBuffer buffer = payload.sizeof() > 0 ? payload.buffer() : null;
+                    final int offset = payload.sizeof() > 0 ? payload.offset() : -1;
+                    final int limit = payload.sizeof() > 0 ? payload.limit() : -1;
```
jfallows: Shouldn't this be a test for `payload != null` instead, passing `null` for `payload` to `doApplicationData` when needed?

With feedback above for `doData` applied, then `payload` would pass directly to `doData` without requiring these `null` checks.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -460,15 +465,21 @@ private void doData(
...
             .flags(flags)
             .budgetId(budgetId)
-            .reserved(reserved)
-            .payload(buffer, index, length)
+            .reserved(reserved);
+
+        if (buffer != null)
+        {
+            builder.payload(buffer, index, length);
+        }
```
jfallows: Can `doData` be refactored to receive `payload` parameter as type `Flyweight` instead of separate `buffer`, `offset` and `limit`, making it easy to pass `null` when needed without the `null` checks only required because we're trying to split it into `buffer`, `offset` and `length`?

jfallows (changes requested):

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
izzymoriguchi: When we get `decodeTransfer` with a null value (`0x40`), the payload won't be null and will be empty octets in the current implementation of `AmqpMessageDecoder` (`messageFragmentRW.build()` is always called so it won't be null).

jfallows: Yes, but by the time it arrives here in `doApplicationData`, doesn't the payload represent the `DATA` frame payload, which means it should already be `null` before being passed?

izzymoriguchi: Yes, that sounds right.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
jfallows: Perhaps change this to:
```
builder.payload(payload != null ? octetsRO.wrap(payload.buffer(), payload.offset(), payload.length()) : null);
```

Then we can go back to the style of code before these changes, where the builder did not need to be extracted separately.

As an aside, perhaps we should consider adding `Flyweight.Builder.wrap(flyweight)` and `Flyweight.Builder.tryWrap(flyweight)`, where `tryWrap` returns `null` if `flyweight` is `null`, then the code here would become:
```
builder.payload(octetsRO.tryWrap(payload));
```

Alternatively, we could generate an overload of `DataFW.Builder.payload(Flyweight payload)` that has the same implementation as `DataFW.Builder.payload(OctetsFW payload)`, then the code would become:
```
builder.payload(payload);
```
This last option seems best to me, what do you think?

jfallows (changes requested):

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
izzymoriguchi: The last option looks cleaner than the other two.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -1152,7 +1157,10 @@ private void doEncodeTransfer(
 
             frameBuffer.putBytes(transfer.limit(), buffer, offset, length);
             replyBudgetReserved += frameSize + replyPadding;
-            doNetworkData(traceId, authorization, 0L, frameBuffer, 0, frameSize);
+            OctetsFW payload = payloadRW.wrap(extraBuffer, 0, extraBuffer.capacity())
+                .set(frameBuffer, 0, frameSize)
+                .build();
```
jfallows: No need to make an extra copy, right?
Suggest:
```
OctetsFW payload = payloadRO.wrap(frameBuffer, 0, frameSize);
```

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -3325,6 +3337,20 @@ private int encodeSectionValue(
...
         }
 
+        private int encodeSectionValueNull(
+            int deferred,
+            DirectBuffer buffer,
+            int offset,
+            int limit)
+        {
+            AmqpDescribedTypeFW descriptor =
+                amqpDescribedTypeRW.wrap(valueBuffer, 0, valueBuffer.capacity()).set(VALUE).build();
```
jfallows: Suggest pre-creating and reusing `VALUE` descriptor instead of making extra copies.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -2936,10 +2946,9 @@ private OctetsFW encodeSections(
             OctetsFW payload)
         {
             AmqpSectionEncoder previous = null;
-            DirectBuffer buffer = payload.buffer();
-            int offset = payload.offset();
-            int progress = offset;
-            int limit = payload.limit();
+            DirectBuffer buffer = payload != null ? payload.buffer() : null;
+            int progress = payload != null ? payload.offset() : -1;
+            int limit = payload != null ? payload.limit() : -1;
```
jfallows: Suggest not attempting to handle `null` case in generic codepath.
Instead put `assert payload != null` here, and then in `encodeFragmentInit` above, consider changing
```
            encodeMessageProperties(dataEx.properties());
            encodeMessageAnnotations(dataEx.annotations());
            encodeApplicationProperties(dataEx.applicationProperties());
            this.sectionEncoder = lookupBodyEncoder(bodyKind);
            return encodeSections(deferred, payload);
```
to
```
            encodeMessageProperties(dataEx.properties());
            encodeMessageAnnotations(dataEx.annotations());
            encodeApplicationProperties(dataEx.applicationProperties());
            if (payload == null)
            {
                return encodeSectionValueNull(deferred, payload);
            }
            else
            {
                this.sectionEncoder = lookupBodyEncoder(bodyKind);
                return encodeSections(deferred, payload);
            }
```
You could then inline `encodeSectionValueNull` instead, and return a pre-combined buffer with `VALUE` descriptor followed by `NULL` value rather than copying them into a combined buffer each time.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -252,6 +254,11 @@
             .set(DATA)
             .build();
 
+    private final OctetsFW nullConstructor = new OctetsFW.Builder()
+        .wrap(new UnsafeBuffer(new byte[1]), 0, 1)
+        .set(new byte[] {0x40})
+        .build();
+
```
jfallows, line 266: This can be simplified to a readonly wrap (instead of an allocate + copy) via:
```
    private final OctetsFW nullConstructor = new OctetsFW()
        .wrap(new UnsafeBuffer(new byte[] {0x40}), 0, 1);
```

jfallows (changes requested):

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
izzymoriguchi: When I changed the code to the suggested way, I get an empty octet? (`oxtets[0]`)

jfallows:
```
    public static void main(String[] args)
    {
        OctetsFW nullConstructor = new OctetsFW()
                .wrap(new UnsafeBuffer(new byte[] {0x40}), 0, 1);

        System.out.println(nullConstructor);
    }
```
outputs
```
octets[1]
```

jfallows (dismissed):

jfallows (approved):

PR #37, opened 2020-09-04
Implement sasl exchange

### LINKED ISSUE #33: Support SASL anonymous mechanism
See AMQP 1.0 specification, [section 5.3](http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-security-v1.0-os.html#section-sasl).

We initially need to support `sasl.init` with `ANONYMOUS` mechanism, and `sasl.outcome` with `ok` code.

izzymoriguchi:
Fix #33.
Requires https://github.com/reaktivity/nukleus-amqp.spec/pull/24.

### DIFF
--- a/src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
+++ b/src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
@@ -167,8 +167,7 @@ public final class AmqpServerFactory implements StreamFactory
     private static final int TRANSFER_HEADER_SIZE = 20;
     private static final int PAYLOAD_HEADER_SIZE = 205;
     private static final int NO_DELIVERY_ID = -1;
-    private static final int DEFAULT_PROTOCOL_ID = 0;
-    private static final int TLS_PROTOCOL_ID = 2;
+    private static final int PLAIN_PROTOCOL_ID = 0;
     private static final int SASL_PROTOCOL_ID = 3;
     private static final long PROTOCOL_HEADER = 0x414D5150_00010000L;
     private static final long PROTOCOL_HEADER_SASL = 0x414D5150_03010000L;
@@ -282,10 +281,10 @@ public final class AmqpServerFactory implements StreamFactory
         .set(VALUE)
         .build();
 
-    private final AmqpProtocolHeaderFW defaultProtocolHeader = new AmqpProtocolHeaderFW.Builder()
+    private final AmqpProtocolHeaderFW plainProtocolHeader = new AmqpProtocolHeaderFW.Builder()
         .wrap(new UnsafeBuffer(new byte[8]), 0, 8)
         .name(n -> n.set("AMQP".getBytes(StandardCharsets.US_ASCII)))
-        .id(DEFAULT_PROTOCOL_ID)
+        .id(PLAIN_PROTOCOL_ID)
         .major(1)
         .minor(0)
         .revision(0)
@@ -330,7 +329,7 @@ public final class AmqpServerFactory implements StreamFactory
     private final BudgetCreditor creditor;
     private final LongFunction<BudgetDebitor> supplyDebitor;
 
-    private final AmqpServerDecoder decodeFrame = this::decodeFrame;
+    private final AmqpServerDecoder decodePlainFrame = this::decodePlainFrame;
     private final AmqpServerDecoder decodeSaslFrame = this::decodeSaslFrame;
     private final AmqpServerDecoder decodeProtocolHeader = this::decodeProtocolHeader;
     private final AmqpServerDecoder decodeProtocolHeaderZero = this::decodeProtocolHeaderZero;
@@ -668,7 +667,7 @@ public final class AmqpServerFactory implements StreamFactory
             int limit);
     }
 
-    private int decodeFrame(
+    private int decodePlainFrame(
         AmqpServer server,
         final long traceId,
         final long authorization,
@@ -695,7 +694,7 @@ public final class AmqpServerFactory implements StreamFactory
             if (frameSize > server.decodeMaxFrameSize)
             {
                 server.onDecodeError(traceId, authorization, CONNECTION_FRAMING_ERROR);
-                server.decoder = decodeFrame;
+                server.decoder = decodePlainFrame;
                 progress = limit;
                 break decode;
             }
@@ -733,7 +732,7 @@ public final class AmqpServerFactory implements StreamFactory
         decode:
         if (length != 0)
         {
-            if (server.isSaslOutcome)
+            if (server.hasSaslOutcome)
             {
                 server.decoder = decodeProtocolHeaderZero;
                 break decode;
@@ -750,7 +749,7 @@ public final class AmqpServerFactory implements StreamFactory
             if (frameSize > server.decodeMaxFrameSize)
             {
                 server.onDecodeError(traceId, authorization, CONNECTION_FRAMING_ERROR);
-                server.decoder = decodeFrame;
+                server.decoder = decodePlainFrame;
                 progress = limit;
                 break decode;
             }
@@ -786,12 +785,9 @@ public final class AmqpServerFactory implements StreamFactory
 
         switch (protocolId)
         {
-        case DEFAULT_PROTOCOL_ID:
+        case PLAIN_PROTOCOL_ID:
             server.onDecodeProtocolHeader(traceId, authorization, protocolHeader);
-            server.decoder = decodeFrame;
-            break;
-        case TLS_PROTOCOL_ID:
-            // TODO: support TLS
+            server.decoder = decodePlainFrame;
             break;
         case SASL_PROTOCOL_ID:
             server.onDecodeSaslProtocolHeader(traceId, authorization, protocolHeader);
@@ -821,9 +817,9 @@ public final class AmqpServerFactory implements StreamFactory
         if (protocolHeader != null)
         {
             final int protocolId = protocolHeader.id();
-            server.decoder = decodeFrame;
+            server.decoder = decodePlainFrame;
 
-            if (protocolId != 0)
+            if (protocolId != PLAIN_PROTOCOL_ID)
             {
                 server.onDecodeError(traceId, authorization, NOT_ALLOWED);
                 progress = limit;
@@ -852,7 +848,7 @@ public final class AmqpServerFactory implements StreamFactory
         // TODO: verify decodeChannel == 0
         // TODO: verify not already open
         server.onDecodeOpen(traceId, authorization, open);
-        server.decoder = decodeFrame;
+        server.decoder = decodePlainFrame;
         return open.limit();
     }
 
@@ -870,7 +866,7 @@ public final class AmqpServerFactory implements StreamFactory
         assert begin != null;
 
         server.onDecodeBegin(traceId, authorization, begin);
-        server.decoder = decodeFrame;
+        server.decoder = decodePlainFrame;
         return begin.limit();
     }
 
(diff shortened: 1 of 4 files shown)


src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -307,6 +322,11 @@
         decodersByPerformative.put(DETACH, decodeDetach);
         decodersByPerformative.put(END, decodeEnd);
         decodersByPerformative.put(CLOSE, decodeClose);
+        // decodersByPerformative.put(SASL_MECHANISMS, decodeSaslMechanisms);
+        decodersByPerformative.put(SASL_INIT, decodeSaslInit);
+        // decodersByPerformative.put(SASL_CHALLENGE, decodeSaslChallenge);
+        // decodersByPerformative.put(SASL_RESPONSE, decodeSaslResponse);
+        // decodersByPerformative.put(SASL_OUTCOME, decodeSaslOutcome);
```
jfallows: `SASL` frames are not the same as `Performatives`.

Instead we need to have a `decodeFrame` and `decodeSaslFrame`, selected based on `protocol-id` in `decodeProtocolHeader`.

Suggest updating `decodeProtocolHeader` such that `protocol-id` of `0` gives `decodeFrame` and `3` gives `decodeSaslFrame` as next decoder.

Suggest adding `decodeProtocolHeaderZero` which requires `protocol-id` of `0`, this would be used after `sasl` exchange is complete to ensure that any other `protocol-id` is rejected at that point.

Suggest keeping `SASL` frame decoders in separate map, perhaps called `decodersBySaslType` that is used in `decodeSaslFrame` to hand off to `decodeSaslInit`.

Suggest renaming `onDecodeHeader` to `onDecodeProtocolHeader` and calling from `decodeProtocolHeader` when `protocol-id` is `0`, or calling `onDecodeSaslProtocolHeader` from `decodeProtocolHeader` when `protocol-id` is `3`, which would then call `doEncodeSaslProtocolHeader`, passing the supported mechanisms as a parameter, currently `ANONYMOUS`, rather than attempting to reuse `doEncodeProtocolHeader` for both non-SASL and SASL cases.

src/main/reaktivity/protocol.idl
```diff
@@ -657,6 +657,43 @@ scope protocol
             AmqpError error;
         }
 
+        list AmqpSaslMechanisms using AmqpList
+        {
+            required AmqpArray<AmqpSymbol> saslServerMechanisms;
```
jfallows: Suggest renaming this field `mechanisms`.

jfallows (changes requested):

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -695,12 +754,45 @@ private int decodeProtocolHeader(
...
+        {
+            server.onDecodeProtocolHeader(traceId, authorization, protocolHeader);
+            server.decoder = decodeFrame;
+        }
+        else if (protocolId == SASL_PROTOCOL_ID)
+        {
+            server.onDecodeSaslProtocolHeader(traceId, authorization, protocolHeader);
+            server.decoder = decodeSaslFrame;
+        }
+        progress = protocolHeader.limit();
```
jfallows, line 799: What if we are given a `protocol-id` of something other than `0` or `3` ?
Suggest defining a `DEFAULT_PROTOCOL_ID` constant with value `0`.
Suggest using a `switch` here instead of `if, else-if`.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -695,12 +754,45 @@ private int decodeProtocolHeader(
...
+    {
+        final AmqpProtocolHeaderFW protocolHeader = amqpProtocolHeaderRO.tryWrap(buffer, offset, limit);
         int progress = offset;
+
         if (protocolHeader != null)
         {
-            server.onDecodeHeader(traceId, authorization, protocolHeader);
-            server.decoder = decodeFrameType;
+            final int protocolId = protocolHeader.id();
+            assert protocolId == 0;
```
jfallows: What if we are given a `protocol-id` of something other than `0` after the `sasl` handshake is complete?

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -918,12 +1010,12 @@ private int decodeSaslInit(
...
         int progress;
-        final AmqpPerformativeFW performative = amqpPerformativeRO.wrap(buffer, offset, limit);
-        final AmqpSaslInitFW saslInit = performative.saslInit();
+        final AmqpSecurityFW security = amqpSecurityRO.wrap(buffer, offset, limit);
+        final AmqpSaslInitFW saslInit = security.saslInit();
         assert saslInit != null;
 
         server.onDecodeSaslInit(traceId, authorization, saslInit);
-        server.decoder = decodeProtocolHeader;
+        server.decoder = decodeProtocolHeaderZero;
```
jfallows: This direct assignment to `decodeProtocolHeaderZero` assumes that the `sasl` handshake is over when `onDecodeSaslInit` returns.

However, even though that is true for the current simplified case, where the `sasl` `outcome` is produced immediately by `onDecodeSaslInit`, that is not the general case where multiple `sasl` frames must interplay back and forth before the `sasl` handshake is complete.

Therefore, it might be better to track the `sasl` `outcome` in `onDecodeSaslInit` just prior to calling `doEncodeSaslOutcome`, have this `decodeSaslInit` always set the decoder back to `decodeSaslFrame` and then in `decodeSaslFrame`, check for `sasl` `outcome`, deferring to `decodeProtocolHeaderZero` if `sasl` handshake is complete.

That subtle change in logic should address both this simplified `sasl` handshake and also the more general case later.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -1030,35 +1122,51 @@ private void doEncodeProtocolHeader(
...
             long authorization)
+        {
+            final AmqpProtocolHeaderFW protocolHeader = amqpProtocolHeaderRW
+                .wrap(writeBuffer, DataFW.FIELD_OFFSET_PAYLOAD, writeBuffer.capacity())
+                .name(n -> n.set("AMQP".getBytes(StandardCharsets.US_ASCII)))
+                .id(protocolId)
+                .major(major)
+                .minor(minor)
+                .revision(revision)
+                .build();
```
jfallows: Seems that this could be pre-created and shared at factory scope rather than rebuilding each time.

jfallows (changes requested):

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
izzymoriguchi: Wouldn't it be an error in that case?

jfallows: Agreed. That implies we should respond in error to the client, rather than asserting, because a failed assert could adversely impact all connections.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -759,15 +784,21 @@ private int decodeProtocolHeader(
...
+        case DEFAULT_PROTOCOL_ID:
             server.onDecodeProtocolHeader(traceId, authorization, protocolHeader);
             server.decoder = decodeFrame;
-        }
-        else if (protocolId == SASL_PROTOCOL_ID)
-        {
+            break;
+        case TLS_PROTOCOL_ID:
+            // TODO: support TLS
+            break;
```
jfallows, line 791: We need to remove this `case`, since otherwise we'd be ignoring `protocol-id` `2`, instead we should be rejecting it, at least for now.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -759,15 +784,21 @@ private int decodeProtocolHeader(
         final int protocolId = protocolHeader.id();
         int progress;
 
-        if (protocolId == 0)
+        switch (protocolId)
         {
+        case DEFAULT_PROTOCOL_ID:
```
jfallows: We might also consider calling this `PLAIN_PROTOCOL_ID` and renaming `decodeFrame` to `decodePlainFrame` as a peer of `decodeSaslFrame`.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -786,13 +817,20 @@ private int decodeProtocolHeaderZero(
...
         int progress = offset;
 
+        decode:
         if (protocolHeader != null)
         {
             final int protocolId = protocolHeader.id();
-            assert protocolId == 0;
+            server.decoder = decodeFrame;
+
+            if (protocolId != 0)
```
jfallows: Use constant instead of hardcoding `0` here.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -1082,6 +1120,8 @@ private int decodeUnknownType(
         private long decodeMaxFrameSize = MIN_MAX_FRAME_SIZE;
         private long encodeMaxFrameSize = MIN_MAX_FRAME_SIZE;
 
+        private boolean isSaslOutcome;
```
jfallows: Suggest renaming to `hasSaslOutcome`.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -1101,48 +1141,23 @@ private AmqpServer(
...
-                .id(protocolId)
-                .major(major)
-                .minor(minor)
-                .revision(revision)
-                .build();
-
-            replyBudgetReserved += protocolHeader.sizeof() + replyPadding;
-            doNetworkData(traceId, authorization, 0L, protocolHeader);
+            replyBudgetReserved += defaultProtocolHeader.sizeof() + replyPadding;
+            doNetworkData(traceId, authorization, 0L, defaultProtocolHeader);
```
jfallows: `defaultProtocolHeader` would name to `plainProtocolHeader`.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -1976,7 +1991,7 @@ private void onDecodeProtocolHeader(
         {
             if (isProtocolHeaderValid(header))
             {
-                doEncodeProtocolHeader(header.id(), header.major(), header.minor(), header.revision(), traceId, authorization);
+                doEncodeProtocolHeader(traceId, authorization);
```
jfallows: Perhaps rename to `doEncodePlainProtocolHeader`.

jfallows (changes requested):

jfallows (dismissed):

jfallows (approved):

PR #38, opened 2020-09-10
Implement idle-timeout

### LINKED ISSUE #34: Support idle timeout
See AMQP 1.0 specification, [section 2.4.5](http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-transport-v1.0-os.html#doc-doc-idle-time-out).

 - detect peer `idle-time-out` for incoming `OPEN`, send empty frame on outbound if idle to prevent timeout (ms)
 - send (configurable) `idle-time-out` for outgoing `OPEN`, teardown if inbound idle for timeout period (ms)

izzymoriguchi:
Fix #34.
Requires https://github.com/reaktivity/nukleus-amqp.spec/pull/26

### DIFF
--- a/src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
+++ b/src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
@@ -174,8 +174,8 @@ public final class AmqpServerFactory implements StreamFactory
     private static final int SASL_PROTOCOL_ID = 3;
     private static final long PROTOCOL_HEADER = 0x414D5150_00010000L;
     private static final long DEFAULT_IDLE_TIMEOUT = 0;
-    private static final int DECODE_IDLE_SIGNAL_ID = 0;
-    private static final int ENCODE_IDLE_SIGNAL_ID = 1;
+    private static final int READ_IDLE_SIGNAL_ID = 0;
+    private static final int WRITE_IDLE_SIGNAL_ID = 1;
     private static final int MIN_IDLE_TIMEOUT = 100;
     private static final long PROTOCOL_HEADER_SASL = 0x414D5150_03010000L;
 
@@ -731,7 +731,7 @@ public final class AmqpServerFactory implements StreamFactory
             server.decoder = decoder;
             if (descriptor != OPEN)
             {
-                server.doSignalEncodeIdleTimeoutIfNecessary();
+                server.doSignalWriteIdleTimeoutIfNecessary();
             }
             progress = performative.offset();
         }
@@ -1138,14 +1138,14 @@ public final class AmqpServerFactory implements StreamFactory
         private long decodableBodyBytes;
         private long decodeMaxFrameSize = MIN_MAX_FRAME_SIZE;
         private long encodeMaxFrameSize = MIN_MAX_FRAME_SIZE;
-        private long encodeIdleTimeout = DEFAULT_IDLE_TIMEOUT;
-        private long decodeIdleTimeout = DEFAULT_IDLE_TIMEOUT;
+        private long writeIdleTimeout = DEFAULT_IDLE_TIMEOUT;
+        private long readIdleTimeout = DEFAULT_IDLE_TIMEOUT;
 
-        private long decodeIdleTimeoutId = NO_CANCEL_ID;
-        private long decodeIdleTimeoutAt;
+        private long readIdleTimeoutId = NO_CANCEL_ID;
+        private long readIdleTimeoutAt;
 
-        private long encodeIdleTimeoutId = NO_CANCEL_ID;
-        private long encodeIdleTimeoutAt;
+        private long writeIdleTimeoutId = NO_CANCEL_ID;
+        private long writeIdleTimeoutAt;
 
         private boolean hasSaslOutcome;
 
@@ -1446,7 +1446,7 @@ public final class AmqpServerFactory implements StreamFactory
                         fragmentBuffer, fragmentProgress, fragmentSizeCont);
                 fragmentProgress += fragmentSizeCont;
                 fragmentRemaining -= fragmentSizeCont;
-                doSignalDecodeIdleTimeoutIfNecessary();
+                doSignalReadIdleTimeoutIfNecessary();
             }
 
             AmqpTransferFW.Builder transferFinBuilder = amqpTransferRW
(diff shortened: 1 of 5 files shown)


src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -160,6 +163,9 @@
     private static final int PAYLOAD_HEADER_SIZE = 205;
     private static final int NO_DELIVERY_ID = -1;
     private static final long PROTOCOL_HEADER = 0x414D5150_00010000L;
+    private static final long DEFAULT_IDLE_TIMEOUT = 0;
+    private static final int DECODE_IDLE_TIMEOUT_ID = 0;
+    private static final int ENCODE_IDLE_TIMEOUT_ID = 1;
```
jfallows: Suggest renaming these to `DECODE_IDLE_SIGNAL_ID` and `ENCODE_IDLE_SIGNAL_ID`.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -267,12 +273,20 @@
     private final OctetsFW nullConstructor = new OctetsFW()
         .wrap(new UnsafeBuffer(new byte[] {0x40}), 0, 1);
 
+
+    private final OctetsFW emptyFrameHeader = new OctetsFW()
+        .wrap(new UnsafeBuffer(new byte[] {0x00, 0x00, 0x00, 0x08, 0x02, 0x00, 0x00, 0x00}), 0, 8);
+
+    private final StringFW timeoutDescription = new String8FW("idle-timeout expired");
+
```
jfallows, line 318: Excellent.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -2001,6 +2108,32 @@ private void cleanupEncodeSlotIfNecessary()
             }
         }
 
+        private void doSignalDecodeIdleTimeoutIfNecessary()
+        {
+            if (decodeIdleTimeout > 0)
+            {
+                decodeIdleTimeoutAt = System.currentTimeMillis() + decodeIdleTimeout;
+
+                if (decodeIdleTimeoutId == NO_CANCEL_ID && decodeIdleTimeout > 0)
```
jfallows: No need for second check of `decodeIdleTimeout > 0`.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -2001,6 +2108,32 @@ private void cleanupEncodeSlotIfNecessary()
...
+                    decodeIdleTimeoutId = signaler.signalAt(decodeIdleTimeoutAt, routeId, replyId, DECODE_IDLE_TIMEOUT_ID);
+                }
+            }
+        }
+
+        private void doSignalEncodeIdleTimeoutIfNecessary()
+        {
+            if (encodeIdleTimeout > 0)
+            {
+                encodeIdleTimeoutAt = System.currentTimeMillis() + encodeIdleTimeout - (encodeIdleTimeout / 2);
```
jfallows: Suggest we simplify as follows:
```
encodeIdleTimeoutAt = System.currentTimeMillis() + (encodeIdleTimeout >> 1);
```

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -948,6 +967,14 @@ private int decodeUnknownType(
         private long decodableBodyBytes;
         private long decodeMaxFrameSize = MIN_MAX_FRAME_SIZE;
         private long encodeMaxFrameSize = MIN_MAX_FRAME_SIZE;
+        private long encodeIdleTimeout = 0;
+        private long decodeIdleTimeout = 0;
```
jfallows: Suggest assigning these to `DEFAULT_IDLE_TIMEOUT` instead of `0` for readability.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -1002,6 +1036,11 @@ private void doEncodeOpen(
                 builder.maxFrameSize(decodeMaxFrameSize);
             }
 
+            if (defaultIdleTimeout != DEFAULT_IDLE_TIMEOUT)
+            {
+                builder.idleTimeOut(defaultIdleTimeout);
+            }
+
```
jfallows, line 1255: Excellent.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -1793,7 +1888,11 @@ private void onDecodeOpen(
             // TODO: use buffer slot capacity instead
             this.encodeMaxFrameSize = Math.min(replySharedBudget, open.maxFrameSize());
             this.decodeMaxFrameSize = defaultMaxFrameSize;
+            this.encodeIdleTimeout = open.hasIdleTimeOut() ? open.idleTimeOut() : 0;
```
jfallows: Please use `DEFAULT_IDLE_TIMEOUT` instead of `0` for readability.

Suggest we enforce a minimum value for `open.idleTimeOut()`, meaning a maximum frequency for empty frames to be sent by the server back to the client.

> If the receiver is unable or unwilling to support the idle time-out then it SHOULD close the connection with an error explaining why (e.g., because it is too small).

Note: in this error case, i think we would initially send the `OPEN` then follow up with an immediate `CLOSE` with explanatory error.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -1828,10 +1929,11 @@ private void onDecodeAttach(
             if (session != null)
             {
                 session.onDecodeAttach(traceId, authorization, attach);
+                doSignalDecodeIdleTimeoutIfNecessary();
```
jfallows: Please update decode idle timeout signal generically during `decodeFrameType` / `decodePlainFrame` instead, much like we enforce maximum frame size.

src/test/java/org/reaktivity/nukleus/amqp/internal/streams/server/AmqpServerIT.java
```diff
@@ -1266,4 +1266,33 @@ public void shouldCloseConnectionWhenMaxFrameSizeExceededWithMultipleSessions()
     {
         k3po.finish();
     }
+
+    @Test
+    @Specification({
+        "${route}/server/controller",
+        "${client}/connection/server.idle.timeout.expires/client" })
+    @Configure(name = "nukleus.amqp.idle.timeout", value = "10000")
```
jfallows: Recommend reducing these timeouts significantly to avoid making the build take a lot of extra time.

src/test/java/org/reaktivity/nukleus/amqp/internal/streams/server/AmqpServerIT.java
```diff
@@ -1266,4 +1266,33 @@ public void shouldCloseConnectionWhenMaxFrameSizeExceededWithMultipleSessions()
...
+    public void shouldCloseConnectionWithTimeout() throws Exception
+    {
+        k3po.finish();
+    }
+
+    @Test
+    @Specification({
+        "${route}/server/controller",
+        "${client}/connection/server.idle.timeout.does.not.expire/client" })
+    @Configure(name = "nukleus.amqp.idle.timeout", value = "10000")
```
jfallows: Recommend reducing these timeouts significantly to avoid making the build take a lot of extra time.

src/test/java/org/reaktivity/nukleus/amqp/internal/streams/server/AmqpServerIT.java
```diff
@@ -39,7 +39,7 @@
         .addScriptRoot("client", "org/reaktivity/specification/amqp")
         .addScriptRoot("server", "org/reaktivity/specification/nukleus/amqp/streams");
 
-    private final TestRule timeout = new DisableOnDebug(new Timeout(10, SECONDS));
+    private final TestRule timeout = new DisableOnDebug(new Timeout(30, SECONDS));
```
jfallows: This should be able to revert to `10 seconds` after making some changes below.

jfallows (changes requested):

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
izzymoriguchi: If a `OPEN` frame doesn't have a `idleTimeout` (`!open.hasIdleTimeOut()`), do we still send the error? In that case, I need to update (add `idleTimeout` field) other spec scripts with `OPEN` frame.

jfallows: According to spec
>The idle timeout REQUIRED by the sender (see subsection 2.4.5). A value of zero is the same as if it was not set (null).

This would suggest that sending `0` amounts to not sending a required field, which would be a protocol error, agree?

jfallows: Actually, the spec is misleading - seems more like the `idle-timeout` behavior is _required_, but only when the `idle-timeout` field of `open` is specified and non-zero.

jfallows (approved):

PR #77, opened 2020-09-23
Implement CLOSE timeout

### LINKED ISSUE #42: Support CLOSE timeout
See AMQP 1.0 specification, [section 2.4.3](http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-transport-v1.0-os.html#doc-idp162944).

Implement timeout after sending `CLOSE` if `CLOSE` not received.

izzymoriguchi:
Fix #42.
Requires https://github.com/reaktivity/nukleus-amqp.spec/pull/30.

### DIFF
--- a/src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
+++ b/src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
@@ -15,6 +15,7 @@
  */
 package org.reaktivity.nukleus.amqp.internal.stream;
 
+import static java.lang.System.currentTimeMillis;
 import static java.nio.ByteOrder.BIG_ENDIAN;
 import static java.nio.charset.StandardCharsets.UTF_8;
 import static java.util.Objects.requireNonNull;
@@ -1868,7 +1869,7 @@ public final class AmqpServerFactory implements StreamFactory
             final long traceId = signal.traceId();
             final long authorization = signal.authorization();
 
-            final long now = System.currentTimeMillis();
+            final long now = currentTimeMillis();
             if (now >= readIdleTimeoutAt)
             {
                 onDecodeError(traceId, authorization, RESOURCE_LIMIT_EXCEEDED, timeoutDescription);
@@ -1886,7 +1887,7 @@ public final class AmqpServerFactory implements StreamFactory
             final long traceId = signal.traceId();
             final long authorization = signal.authorization();
 
-            final long now = System.currentTimeMillis();
+            final long now = currentTimeMillis();
             if (now >= writeIdleTimeoutAt)
             {
                 writeIdleTimeoutId = NO_CANCEL_ID;
@@ -1927,8 +1928,7 @@ public final class AmqpServerFactory implements StreamFactory
             StringFW errorDescription)
         {
             cleanupStreams(traceId, authorization);
-            doEncodeClose(traceId, authorization, errorType, errorDescription);
-            doNetworkEndIfNecessary(traceId, authorization);
+            doEncodeCloseAndEndIfNecessary(traceId, authorization, errorType, errorDescription);
         }
 
         private void doNetworkBegin(
@@ -2253,8 +2253,7 @@ public final class AmqpServerFactory implements StreamFactory
             AmqpCloseFW close)
         {
             sessions.values().forEach(s -> s.cleanup(traceId, authorization));
-            doEncodeCloseIfNecessary(traceId, authorization);
-            doNetworkEndIfNecessary(traceId, authorization);
+            doEncodeCloseAndEndIfNecessary(traceId, authorization, null, null);
             doCancelCloseTimeoutIfNecessary();
         }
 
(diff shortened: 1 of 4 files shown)


src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -2367,6 +2380,16 @@ private void doSignalWriteIdleTimeoutIfNecessary()
             }
         }
 
+        private void doSignalCloseTimeout()
+        {
+            closeTimeoutAt = System.currentTimeMillis() + CLOSE_TIMEOUT;
```
jfallows: This can be a local variable, no need to hold on to it in instance state, agree?

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -2367,6 +2380,16 @@ private void doSignalWriteIdleTimeoutIfNecessary()
...
+
+            if (closeTimeoutId == NO_CANCEL_ID)
+            {
+                closeTimeoutId = signaler.signalAt(closeTimeoutAt, routeId, replyId, CLOSE_SIGNAL_ID);
+            }
```
jfallows: Under what conditions do we need to make this check?
If it is expected to always be true it can be changed to an assert.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -2367,6 +2380,16 @@ private void doSignalWriteIdleTimeoutIfNecessary()
             }
         }
 
+        private void doSignalCloseTimeout()
+        {
+            closeTimeoutAt = System.currentTimeMillis() + CLOSE_TIMEOUT;
+
+            if (closeTimeoutId == NO_CANCEL_ID)
+            {
+                closeTimeoutId = signaler.signalAt(closeTimeoutAt, routeId, replyId, CLOSE_SIGNAL_ID);
```
jfallows: We should cancel this `closeTimeoutId` if not  equal to `NO_CANCEL_ID` when `CLOSE` frame is decoded, agree?

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -176,6 +176,8 @@
     private static final long DEFAULT_IDLE_TIMEOUT = 0;
     private static final int READ_IDLE_SIGNAL_ID = 0;
     private static final int WRITE_IDLE_SIGNAL_ID = 1;
+    private static final int CLOSE_SIGNAL_ID = 2;
+    private static final int CLOSE_TIMEOUT = 500;
```
jfallows: Please shift `CLOSE_TIMEOUT` to `AmqpConfiguration.CLOSE_EXCHANGE_TIMEOUT`, defaulted to `10000ms`, then override in IT to keep build time down while still validating state machine behavior.

jfallows (changes requested):

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
izzymoriguchi: yes

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -1911,7 +1928,7 @@ private void onDecodeError(
         {
             cleanupStreams(traceId, authorization);
             doEncodeClose(traceId, authorization, errorType, errorDescription);
-            doNetworkEnd(traceId, authorization);
+            doNetworkEndIfNecessary(traceId, authorization);
```
jfallows: If network end is conditional, then don't we need to also guard against encoding `CLOSE` after already ended?

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -2281,6 +2296,16 @@ private void cleanupStreams(
...
         }
 
+        private void doEncodeCloseIfNecessary(
+            long traceId,
+            long authorization)
+        {
+            if (!AmqpState.replyClosed(state))
+            {
+                doEncodeClose(traceId, authorization, null, null);
+            }
```
jfallows, line 2308: This doesn't seem quite right because generally we would expect the state guard to be flipped by the code inside the if check, such that calling it a second time would be a no-op, but that isn't the case here because the state is only flipped when we end the stream.

Perhaps we need a combined `doEncodeCloseAndEndIfNecessary` which would look like above but with `doNetworkEnd(traceId, authorization);` inside the if check, guaranteeing that the state will be flipped as needed and a subsequent call to `doEncodeCloseAndEndIfNecessary` would be a no-op?

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -2367,6 +2401,14 @@ private void doSignalWriteIdleTimeoutIfNecessary()
             }
         }
 
+        private void doSignalCloseTimeout()
+        {
+            long closeTimeoutAt = System.currentTimeMillis() + closeTimeout;
```
jfallows: `final`, also static import for `currentTimeMillis()`.

jfallows (changes requested):

jfallows (dismissed):

jfallows (approved):

PR #80, opened 2020-09-25
Implement connection state tracking

### LINKED ISSUE #44: Implement connection state tracking
See AMQP 1.0 specification, [section 2.4.6](http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-transport-v1.0-os.html#doc-idp184736).

Suggest an `enum` for connection state that has methods indicating what frame has been received, then returning the new connection state, possibly an error state for an invalid transition that causes a protocol error to be sent back to the client.

izzymoriguchi:
Fix #44

### DIFF
--- a/src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
+++ b/src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
@@ -1196,7 +1196,7 @@ public final class AmqpServerFactory implements StreamFactory
             if (!hasSaslOutcome)
             {
                 doEncodePlainProtocolHeader(traceId, authorization);
-                connectionState = AmqpConnectionState.sentHeader(connectionState);
+                connectionState = connectionState.sentHeader();
                 assert connectionState != ERROR && connectionState == HDR_EXCH;
             }
         }
@@ -2123,7 +2123,7 @@ public final class AmqpServerFactory implements StreamFactory
             long authorization,
             AmqpProtocolHeaderFW header)
         {
-            connectionState = AmqpConnectionState.receivedHeader(connectionState);
+            connectionState = connectionState.receivedHeader();
 
             if (connectionState != ERROR && (connectionState == HDR_RCVD || connectionState == OPEN_SENT))
             {
@@ -2135,7 +2135,7 @@ public final class AmqpServerFactory implements StreamFactory
                 else if (!hasSaslOutcome)
                 {
                     doEncodeOpen(traceId, authorization);
-                    connectionState = AmqpConnectionState.sentOpen(connectionState);
+                    connectionState = connectionState.sentOpen();
                     assert connectionState != ERROR && connectionState == OPEN_SENT;
                 }
             }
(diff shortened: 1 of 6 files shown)


src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -472,6 +477,7 @@ private MessageConsumer newInitialStream(
             final long affinity = begin.affinity();
 
             newStream = new AmqpServer(sender, routeId, initialId, affinity)::onNetwork;
+            this.connectionState = START;
```
jfallows: The state should be stored on the `server` since we can be decoding more than one `AmqpServer` in parallel.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -843,7 +855,16 @@ private int decodeProtocolHeaderZero(
...
+            connectionState = connectionState.receivedHeader();
+            if (connectionState != ERROR)
+            {
+                server.onDecodeProtocolHeader(traceId, authorization, protocolHeader);
+            }
+            else
+            {
+                server.onDecodeError(traceId, authorization, ILLEGAL_STATE, null);
+            }
+
```
jfallows, line 860: Recommend you invert the logic, using a `decode:` label.
```
decode:
if (protocolHeader != null)
{
    connectionState = connectionState.receivedHeader();
    if (connectionState == ERROR)
    {
        server.onDecodeError(traceId, authorization, ILLEGAL_STATE, null);
        break decode;
    }

    final int protocolId = protocolHeader.id();
    server.decoder = decodePlainFrame;
    server.onDecodeProtocolHeader(traceId, authorization, protocolHeader);
```

Also, should `server.onDecodeError` be `decodeError` instead to change `decoder` to `decodeIgnoreAll`?

This feedback applies to all places where we detect `ERROR` connection state.

src/test/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpConnectionStateTest.java
```diff
@@ -0,0 +1,170 @@
...
+import org.junit.Test;
+
+public class AmqpConnectionStateTest
+{
+    @Test
+    public void shouldTransitionFromStartToHeaderReceived() throws Exception
+    {
+        AmqpConnectionState state = START.receivedHeader();
+        assertEquals(HDR_RCVD, state);
+    }
```
jfallows, line 48: Suggest you have one test method for each state, called something like `shouldTransitionFromStart()`  and then have assert for all the possible transitions, good and bad.

For example:
```
    @Test
    public void shouldTransitionFromStart() throws Exception
    {
        assertEquals(HDR_SENT, START.sentHeader());
        assertEquals(HDR_RCVD, START.receivedHeader());
        assertEquals(ERROR, START.sentOpen());
        assertEquals(ERROR, START.receivedOpen());
        assertEquals(ERROR, START.sentClose());
        assertEquals(ERROR, START.receivedClose());
    }
```

jfallows (changes requested):

jfallows (dismissed):

jfallows (approved):

jfallows (approved):

PR #82, opened 2020-09-29
Implement session state tracking

### LINKED ISSUE #47: Implement session state tracking
See AMQP 1.0 specification, [section 2.5.5](http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-transport-v1.0-os.html#doc-idp243424).

Suggest an `enum` for session state that has methods indicating what session frame has been received, then returning the new session state, possibly an error state for an invalid transition that causes a protocol error to be sent back to the client.

izzymoriguchi:
Fix #47

### DIFF
--- a/src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
+++ b/src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
@@ -2771,9 +2771,9 @@ public final class AmqpServerFactory implements StreamFactory
                 long traceId,
                 long authorization)
             {
+                links.values().forEach(l -> l.cleanup(traceId, authorization));
                 sessions.remove(incomingChannel);
                 flushReplySharedBudget(traceId);
-                links.values().forEach(l -> l.cleanup(traceId, authorization));
             }
 
             private class AmqpServerStream
--- a/src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
+++ b/src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
@@ -2385,6 +2385,7 @@ public final class AmqpServerFactory implements StreamFactory
                     break decode;
                 }
                 session.doEncodeEnd(traceId, authorization, errorType);
+                session.cleanup(traceId, authorization);
             }
         }
 
@@ -2717,6 +2718,7 @@ public final class AmqpServerFactory implements StreamFactory
                 if (incomingWindow < 0)
                 {
                     doEncodeEnd(traceId, authorization, SESSION_WINDOW_VIOLATION);
+                    cleanup(traceId, authorization);
                 }
                 else
                 {
@@ -2760,9 +2762,6 @@ public final class AmqpServerFactory implements StreamFactory
                 long authorization,
                 AmqpErrorType errorType)
             {
-                cleanup(traceId, authorization);
-                sessions.remove(incomingChannel);
-                flushReplySharedBudget(traceId);
                 AmqpServer.this.doEncodeEnd(traceId, authorization, outgoingChannel, errorType);
                 sessionState = sessionState.sentEnd();
                 assert sessionState != AmqpSessionState.ERROR;
@@ -2772,6 +2771,8 @@ public final class AmqpServerFactory implements StreamFactory
                 long traceId,
                 long authorization)
             {
+                sessions.remove(incomingChannel);
+                flushReplySharedBudget(traceId);
                 links.values().forEach(l -> l.cleanup(traceId, authorization));
             }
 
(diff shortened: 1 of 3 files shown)


src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -2698,6 +2726,8 @@ private void onDecodeTransfer(
                     sessions.remove(incomingChannel);
                     flushReplySharedBudget(traceId);
                     doEncodeEnd(traceId, authorization, outgoingChannel, SESSION_WINDOW_VIOLATION);
+                    sessionState = sessionState.sentEnd();
+                    assert sessionState == AmqpSessionState.DISCARDING;
```
jfallows: Could these be moved to a session-scoped `doEncodeEnd(traceId, authorization)` with fewer arguments instead?
Then it doesn't matter why `END` was sent, the state machine always represents the session state accurately.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -2601,6 +2627,8 @@ private void onDecodeBegin(
                 long authorization)
             {
                 doEncodeBegin(traceId, authorization, incomingChannel, nextOutgoingId);
+                sessionState = sessionState.sentBegin();
+                assert sessionState != AmqpSessionState.ERROR;
```
jfallows: Could these be moved to a session-scoped `doEncodeBegin(traceId, authorization)` with fewer arguments instead?
Then it doesn't matter why `BEGIN` was sent, the state machine always represents the session state accurately.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -2360,12 +2375,20 @@ private void onDecodeEnd(
...
             AmqpSession session = sessions.get(decodeChannel);
+            decode:
             if (session != null)
             {
+                session.sessionState = session.sessionState.receivedEnd();
+                if (session.sessionState == AmqpSessionState.ERROR)
+                {
+                    break decode;
+                }
                 session.cleanup(traceId, authorization);
```
jfallows, line 2388: Suggest handling to `doEncodeEnd` if necessary from inside `session.cleanup(...)` instead.
That will allow the code below, including `doEncodeEnd(...)` and the session state changes to be removed from this method.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -913,14 +914,20 @@ private int decodeBegin(
...
 
+        decode:
         if (performative != null)
         {
             final AmqpBeginFW begin = performative.begin();
             assert begin != null;
 
             server.onDecodeBegin(traceId, authorization, begin);
-            server.decoder = decodePlainFrame;
+            if (server.sessions.get(server.decodeChannel).sessionState == AmqpSessionState.ERROR)
```
jfallows: What if `session` is not found in the `server.sessions` map?

jfallows (changes requested):

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
izzymoriguchi: Do you mean to create a method `doEncodeBegin(traceId, authorization)` in `AmqpSession`, and then move those 3 lines to that method
```
doEncodeBegin(traceId, authorization, incomingChannel, nextOutgoingId);
sessionState = sessionState.sentBegin();
assert sessionState != AmqpSessionState.ERROR;
```
and then call the `doEncodeBegin(traceId, authorization)` method inside this `onDecodeBegin()` method?

jfallows: Sounds good, yes!

jfallows (changes requested):

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
izzymoriguchi: Can I add error type argument in `doEncodeEnd(traceId, authorization)`?  
```
            private void doEncodeEnd(
                long traceId,
                long authorization,
                AmqpErrorType errorType)
            {
                cleanup(traceId, authorization);
                sessions.remove(incomingChannel);
                flushReplySharedBudget(traceId);
                AmqpServer.this.doEncodeEnd(traceId, authorization, outgoingChannel, errorType);
                sessionState = sessionState.sentEnd();
                assert sessionState == AmqpSessionState.DISCARDING;
            }
```

jfallows: Yes, of course!

jfallows (approved):

PR #83, opened 2020-09-30
Implement discarding inbound frame after outout end

### LINKED ISSUE #46: Support discard inbound after outbound END
See AMQP 1.0 specification, [section 2.5.4](http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-transport-v1.0-os.html#doc-idp239696).

When END with error gets sent out, discard all incoming frames until the peer receives the corresponding END.

izzymoriguchi:
Fix #46 
Requires https://github.com/reaktivity/nukleus-amqp.spec/pull/34

### DIFF
--- a/src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
+++ b/src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
@@ -275,42 +275,42 @@ public final class AmqpServerFactory implements StreamFactory
     private final Array8FW.Builder<AmqpSymbolFW.Builder, AmqpSymbolFW> annonymousRW =
         new Array8FW.Builder<>(new AmqpSymbolFW.Builder(), new AmqpSymbolFW());
 
-    private final AmqpPerformativeTypeFW openPerformativeType = new AmqpPerformativeTypeFW.Builder()
+    private final AmqpPerformativeTypeFW openType = new AmqpPerformativeTypeFW.Builder()
         .wrap(new UnsafeBuffer(new byte[3]), 0, 3)
         .set(OPEN)
         .build();
 
-    private final AmqpPerformativeTypeFW beginPerformativeType = new AmqpPerformativeTypeFW.Builder()
+    private final AmqpPerformativeTypeFW beginType = new AmqpPerformativeTypeFW.Builder()
         .wrap(new UnsafeBuffer(new byte[3]), 0, 3)
         .set(BEGIN)
         .build();
 
-    private final AmqpPerformativeTypeFW attachPerformativeType = new AmqpPerformativeTypeFW.Builder()
+    private final AmqpPerformativeTypeFW attachType = new AmqpPerformativeTypeFW.Builder()
         .wrap(new UnsafeBuffer(new byte[3]), 0, 3)
         .set(ATTACH)
         .build();
 
-    private final AmqpPerformativeTypeFW flowPerformativeType = new AmqpPerformativeTypeFW.Builder()
+    private final AmqpPerformativeTypeFW flowType = new AmqpPerformativeTypeFW.Builder()
         .wrap(new UnsafeBuffer(new byte[3]), 0, 3)
         .set(FLOW)
         .build();
 
-    private final AmqpPerformativeTypeFW transferPerformativeType = new AmqpPerformativeTypeFW.Builder()
+    private final AmqpPerformativeTypeFW transferType = new AmqpPerformativeTypeFW.Builder()
         .wrap(new UnsafeBuffer(new byte[3]), 0, 3)
         .set(TRANSFER)
         .build();
 
-    private final AmqpPerformativeTypeFW detachPerformativeType = new AmqpPerformativeTypeFW.Builder()
+    private final AmqpPerformativeTypeFW detachType = new AmqpPerformativeTypeFW.Builder()
         .wrap(new UnsafeBuffer(new byte[3]), 0, 3)
         .set(DETACH)
         .build();
 
-    private final AmqpPerformativeTypeFW endPerformativeType = new AmqpPerformativeTypeFW.Builder()
+    private final AmqpPerformativeTypeFW endType = new AmqpPerformativeTypeFW.Builder()
         .wrap(new UnsafeBuffer(new byte[3]), 0, 3)
         .set(END)
         .build();
 
-    private final AmqpPerformativeTypeFW closePerformativeType = new AmqpPerformativeTypeFW.Builder()
+    private final AmqpPerformativeTypeFW closeType = new AmqpPerformativeTypeFW.Builder()
         .wrap(new UnsafeBuffer(new byte[3]), 0, 3)
         .set(CLOSE)
         .build();
@@ -415,7 +415,7 @@ public final class AmqpServerFactory implements StreamFactory
     private final AmqpServerDecoder decodeClose = this::decodeClose;
     private final AmqpServerDecoder decodeSaslInit = this::decodeSaslInit;
     private final AmqpServerDecoder decodeIgnoreAll = this::decodeIgnoreAll;
-    private final AmqpServerDecoder decodeIgnorePlainFrame = this::decodeIgnorePlainFrame;
+    private final AmqpServerDecoder decodeIgnoreFrameBody = this::decodeIgnoreFrameBody;
     private final AmqpServerDecoder decodeUnknownType = this::decodeUnknownType;
 
     private final int outgoingWindow;
@@ -821,11 +821,12 @@ public final class AmqpServerFactory implements StreamFactory
             final AmqpServer.AmqpSession session = server.sessions.get(server.decodeChannel);
             if (session != null && session.sessionState == AmqpSessionState.DISCARDING && descriptor != END)
             {
-                server.decoder = decodeIgnorePlainFrame;
+                server.decoder = decodeIgnoreFrameBody;
                 break decode;
             }
 
             server.decoder = decodersByPerformativeType.getOrDefault(descriptor, decodeUnknownType);
+            server.decodableBodyBytes -= PERFORMATIVE_SIZE;
             progress = performativeType.limit();
         }
         else
@@ -1107,7 +1108,7 @@ public final class AmqpServerFactory implements StreamFactory
                 {
                     break decode;
                 }
-                server.decodableBodyBytes -= transfer.sizeof() + PERFORMATIVE_SIZE;
+                server.decodableBodyBytes -= transfer.sizeof();
                 final int fragmentOffset = transfer.limit();
                 final int fragmentSize = (int) server.decodableBodyBytes;
                 final int fragmentLimit = fragmentOffset + fragmentSize;
@@ -1268,7 +1269,7 @@ public final class AmqpServerFactory implements StreamFactory
         return limit;
     }
 
-    private int decodeIgnorePlainFrame(
+    private int decodeIgnoreFrameBody(
         AmqpServer server,
         long traceId,
         long authorization,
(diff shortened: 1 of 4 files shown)


src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -736,6 +738,15 @@ private int decodePlainFrame(
             final AmqpDescribedType descriptor = performative.kind();
             final AmqpServerDecoder decoder = decodersByPerformative.getOrDefault(descriptor, decodeUnknownType);
             server.decodeChannel = frameHeader.channel();
+
+            if (server.sessions.get(server.decodeChannel) != null &&
+                server.sessions.get(server.decodeChannel).sessionState == AmqpSessionState.DISCARDING &&
```
jfallows: Please hoist to `session` variable instead of repeating the lookup.

jfallows (changes requested):

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
izzymoriguchi: Can I change `AmqpServer` and `AmqpSession` to package-private from private?

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -736,6 +738,14 @@ private int decodePlainFrame(
             final AmqpDescribedType descriptor = performative.kind();
             final AmqpServerDecoder decoder = decodersByPerformative.getOrDefault(descriptor, decodeUnknownType);
             server.decodeChannel = frameHeader.channel();
+
+            AmqpServer.AmqpSession session = server.sessions.get(server.decodeChannel);
+            if (session != null && session.sessionState == AmqpSessionState.DISCARDING && descriptor != END)
+            {
+                progress = limit;
```
jfallows: Shouldn't we make sure to advance only to the end of the frame?
For example, what if a `TRANSFER` and `END` frame would be sent in the same packet and arrive in the same `DATA` payload?

jfallows (changes requested):

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -1027,7 +1107,7 @@ private int decodeTransfer(
                 {
                     break decode;
                 }
-                server.decodableBodyBytes -= transfer.sizeof();
+                server.decodableBodyBytes -= transfer.sizeof() + PERFORMATIVE_SIZE;
```
jfallows: Why did this change?

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -1194,6 +1268,28 @@ private int decodeIgnoreAll(
...
+        int offset,
+        int limit)
+    {
+        final int length = limit - offset;
+
+        int progress = offset;
+
+        if (length != 0)
+        {
+            progress = (int) (offset + server.decodableBodyBytes);
```
jfallows: This implementation relies on side effects of `decodePlainFrame` preceding it, but it would be better to make this `decodeIgnorePlainFrame` side effect free.

First, it needs to wrap the frame header and verify that the complete frame body is present, only then should it advance the progress to the frame body limit computed from the frame header fields alone.

This current implementation seems susceptible to an incomplete frame body.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -1688,16 +1819,21 @@ private void doEncodeClose(
                 close = builder.build();
             }
 
+            final int totalPayloadSize = FRAME_HEADER_SIZE + PERFORMATIVE_SIZE + close.sizeof();
```
jfallows: Please rename `totalPayloadSize` to `size`.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -265,6 +275,46 @@
     private final Array8FW.Builder<AmqpSymbolFW.Builder, AmqpSymbolFW> annonymousRW =
         new Array8FW.Builder<>(new AmqpSymbolFW.Builder(), new AmqpSymbolFW());
 
+    private final AmqpPerformativeTypeFW openPerformativeType = new AmqpPerformativeTypeFW.Builder()
+        .wrap(new UnsafeBuffer(new byte[3]), 0, 3)
+        .set(OPEN)
+        .build();
```
jfallows, line 281: Suggest we rename this to `openType` for readability, similar feedback for other performative types.

jfallows (changes requested):

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
izzymoriguchi: Because we assign the value of `decodableBodyBytes` as
`server.decodableBodyBytes = frameSize - frameHeader.doff() * 4;`
and this includes performative (3 bytes), `TRANSFER` list and payload.

And so I wanted to subtract the amount of the performative bytes (3 bytes)

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
izzymoriguchi: I'm setting `decoder = decodeIgnorePlainFrame` in `decodePerformative()`. So the `offset` in this `decodeIgnorePlainFrame()` starts from just after the frame header. So I cannot wrap the frame header here.

jfallows: Understood, then let's rename this as `decodeIgnoreFrameBody`.
The point about potentially advancing `progress` beyond `limit` still stands though, so you'll need to handle incrementally ignoring the remaining `server.decodableBytes` until it becomes `0` before updating the decoder to decode the next plain frame.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -732,15 +783,55 @@ private int decodePlainFrame(
...
+
+            final AmqpServer.AmqpSession session = server.sessions.get(server.decodeChannel);
+            if (session != null && session.sessionState == AmqpSessionState.DISCARDING && descriptor != END)
+            {
+                server.decoder = decodeIgnoreFrameBody;
+                break decode;
+            }
+
+            server.decoder = decodersByPerformativeType.getOrDefault(descriptor, decodeUnknownType);
+            server.decodableBodyBytes -= PERFORMATIVE_SIZE;
```
jfallows: Need to add an `assert server.decodableBytes >= 0;`
Also might be better to use `server.decodableBytes -= performativeType.sizeof()`, for example what if it is sent using an 8-byte `ulong` instead of a 1-byte `ulong`?

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -1338,13 +1440,13 @@ private void doEncodeSaslOutcome(
...
             final AmqpSaslOutcomeFW saslOutcome =
-                amqpSaslOutcomeRW.wrap(frameBuffer, FRAME_HEADER_SIZE, frameBuffer.capacity())
+                amqpSaslOutcomeRW.wrap(frameBuffer, FRAME_HEADER_SIZE + PERFORMATIVE_SIZE, frameBuffer.capacity())
                     .code(OK)
                     .build();
 
             final AmqpSaslFrameHeaderFW saslFrameHeader =
                 amqpSaslFrameHeaderRW.wrap(frameBuffer, 0, frameBuffer.capacity())
-                    .size(FRAME_HEADER_SIZE + saslOutcome.sizeof())
+                    .size(FRAME_HEADER_SIZE + PERFORMATIVE_SIZE + saslOutcome.sizeof())
```
jfallows: `sasl` frames don't have a `performative`, that just for `plain` frames, right?

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -1396,23 +1503,28 @@ private void doEncodeBegin(
...
+                .size(size)
                 .doff(2)
                 .type(0)
                 .channel(outgoingChannel)
-                .performative(b -> b.begin(begin))
                 .build();
 
-            replyBudgetReserved += frameHeader.sizeof() + replyPadding;
-            doNetworkData(traceId, authorization, 0L, frameHeader);
+            frameBuffer.putBytes(frameHeader.limit(), beginType.buffer(), 0, PERFORMATIVE_SIZE);
```
jfallows: This should really be `beginType.sizeof()` instead of `PERFORMATIVE_SIZE`.

Suggest reordering this to put the performative first using `FRAME_HEADER_SIZE`, so that it reads as if you are building up the entire frame body, then using that information to fill in the size field of the frame header.

Then after filling in the frame header, add an `assert frameHeader.sizeof() == FRAME_HEADER_SIZE` to ensure your assumptions about positional layout are either always correct, or detected as inconsistent by the assert, perhaps in future if the code is modified.

Please apply this feedback to each of the decodeXXX methods for performatives.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -1556,11 +1683,11 @@ private void doEncodeTransferFragments(
             int fragmentRemaining = fragmentLimit - fragmentProgress;
 
             AmqpTransferFW transferCont = amqpTransferRW
-                    .wrap(frameBuffer, FRAME_HEADER_SIZE, frameBuffer.capacity())
+                    .wrap(frameBuffer, FRAME_HEADER_SIZE + PERFORMATIVE_SIZE, frameBuffer.capacity())
                     .handle(handle)
                     .more(1)
                     .build();
-            int fragmentSizeCont = (int) encodeMaxFrameSize - FRAME_HEADER_SIZE - transferCont.sizeof();
+            int fragmentSizeCont = (int) encodeMaxFrameSize - FRAME_HEADER_SIZE - PERFORMATIVE_SIZE - transferCont.sizeof();
```
jfallows: Given that `replySharedBudget` is an `int`, let's change the type of `encodeMaxFrameSize` to `int` and eliminate these casts.

jfallows (changes requested):

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
izzymoriguchi: well `sasl` frames don't have a performative but it still has 3 byte descriptor

jfallows: Then we need a different constant to make the code more readable please.

jfallows (dismissed):

jfallows (approved):

PR #84, opened 2020-10-01
Support rejecting errant links

### LINKED ISSUE #48: Reject errant links
See AMQP 1.0 specification, [section 2.6.5](http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-transport-v1.0-os.html#doc-idp348080).

Implement an errant-link session error when receiving a handle of a link that has been destroyed.

izzymoriguchi:
Fix #48 
Requires https://github.com/reaktivity/nukleus-amqp.spec/pull/35

### DIFF
--- a/src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
+++ b/src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
@@ -2715,7 +2715,6 @@ public final class AmqpServerFactory implements StreamFactory
         private final class AmqpSession
         {
             private final Long2ObjectHashMap<AmqpServerStream> links;
-            private final Long2ObjectHashMap<AmqpServerStream> destroyedLinks;
             private final int incomingChannel;
 
             private long deliveryId = NO_DELIVERY_ID;
@@ -2734,7 +2733,6 @@ public final class AmqpServerFactory implements StreamFactory
                 int incomingChannel)
             {
                 this.links = new Long2ObjectHashMap<>();
-                this.destroyedLinks = new Long2ObjectHashMap<>();
                 this.incomingChannel = incomingChannel;
                 this.nextOutgoingId++;
                 this.sessionState = UNMAPPED;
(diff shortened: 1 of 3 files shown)


src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -2815,14 +2825,20 @@ private void onDecodeAttach(
...
 
                         AmqpServerStream link = new AmqpServerStream(addressFrom, addressTo, role, route);
-                        AmqpServerStream oldLink = links.put(linkKey, link);
+                        AmqpServerStream oldLink = links.put(handle, link);
                         assert oldLink == null;
+                        if (link.detachError)
+                        {
+                            onDecodeError(traceId, authorization, SESSION_ERRANT_LINK);
+                            break decode;
+                        }
```
jfallows: If the `link` was just created, can `link.detachError` be `true` here?

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -2982,6 +3010,7 @@ private void cleanup(
                     this.initialId = supplyInitialId.applyAsLong(newRouteId);
                     this.replyId = supplyReplyId.applyAsLong(initialId);
                     this.application = router.supplyReceiver(initialId);
+                    this.detachError = false;
```
jfallows: No need to explicitly assign language-defined default values, like `false` for `boolean`.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -2962,6 +2988,8 @@ private void cleanup(
                 private StringFW addressFrom;
                 private StringFW addressTo;
 
+                private boolean detachError;
```
jfallows: Suggest changing the type to `AmqpErrorType` instead of a boolean flag, may be more helpful in debugging.

jfallows (changes requested):

jfallows (dismissed):

jfallows (approved):

PR #85, opened 2020-10-08
Implement transfer with aborted

### LINKED ISSUE #52: Support `abort` of large message transfers
Blocked-By: https://github.com/reaktivity/reaktor.java/issues/65
Blocked-By: https://github.com/reaktivity/reaktor.java/issues/66
Blocked-By: https://github.com/reaktivity/reaktor.java/issues/70

See AMQP 1.0 specification, [section 2.6.14](http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-transport-v1.0-os.html#doc-idp484080).

The sender MAY indicate an aborted attempt to deliver a message by setting the abort flag on the last transfer.

In this case the receiver MUST discard the message data that was transferred prior to the abort.

### LINKED ISSUE #63: Verify `TRANSFER` with `aborted` flag
Blocked-By: https://github.com/reaktivity/reaktor.java/issues/65
Blocked-By: https://github.com/reaktivity/reaktor.java/issues/66
Blocked-By: https://github.com/reaktivity/reaktor.java/issues/70

See AMQP 1.0 specification, [section 2.7.5](http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-transport-v1.0-os.html#type-transfer).

The fields `delivery-id`, `delivery-tag` and `message-format` must be consistent if specified in subsequent `TRANSFER` frames for the same message.

izzymoriguchi:
Fix #52 
Fix #63 
Requires https://github.com/reaktivity/nukleus-amqp.spec/pull/37

### DIFF
--- a/src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
+++ b/src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
@@ -3525,7 +3525,7 @@ public final class AmqpServerFactory implements StreamFactory
                     OctetsFW payload)
                 {
                     final boolean more = (flags & FLAG_FIN) == 0;
-                    final boolean aborted = (flags & FLAG_INIT) == FLAG_INIT;
+                    final boolean aborted = (flags & FLAG_INCOMPLETE) == FLAG_INCOMPLETE;
 
                     OctetsFW messageFragment = amqpMessageHelper.encodeFragment(encodeBodyKind, payload);
 
--- a/src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
+++ b/src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
@@ -191,8 +191,8 @@ public final class AmqpServerFactory implements StreamFactory
     private static final int FLAG_FIN = 1;
     private static final int FLAG_INIT = 2;
     private static final int FLAG_INCOMPLETE = 4;
-    private static final int FLAG_INIT_INCOMPLETE = 6;
-    private static final int FLAG_INIT_AND_FIN = 3;
+    private static final int FLAG_INIT_INCOMPLETE = FLAG_INIT | FLAG_INCOMPLETE;
+    private static final int FLAG_INIT_AND_FIN = FLAG_INIT | FLAG_FIN;
     private static final int FRAME_HEADER_SIZE = 8;
     private static final int SASL_DESCRIPTOR_SIZE = 3;
     private static final int MIN_MAX_FRAME_SIZE = 512;
@@ -3246,6 +3246,7 @@ public final class AmqpServerFactory implements StreamFactory
                     {
                         flags = FLAG_INCOMPLETE;
                     }
+
                     if (!more && !aborted)
                     {
                         flags |= FLAG_FIN;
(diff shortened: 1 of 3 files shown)


src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -3509,6 +3525,7 @@ private void onApplicationDataContOrFin(
                     OctetsFW payload)
                 {
                     final boolean more = (flags & FLAG_FIN) == 0;
+                    final boolean aborted = (flags & FLAG_INIT) == FLAG_INIT;
```
jfallows: This should be `FLAG_INCOMPLETE` instead of `FLAG_INIT`, agreed?
`onApplicationDataInit` should be able to handle a `DATA` frame with both `FLAG_INIT` and `FLAG_INCOMPLETE` set, but there's no reference to `aborted` in there at the moment.

jfallows (changes requested):

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
izzymoriguchi: Yes, sorry I meant to have `FLAG_INCOMPLETE` there.

jfallows (dismissed):

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -190,6 +190,8 @@
 
     private static final int FLAG_FIN = 1;
     private static final int FLAG_INIT = 2;
+    private static final int FLAG_INCOMPLETE = 4;
+    private static final int FLAG_INIT_INCOMPLETE = 6;
```
jfallows: Suggest making this more readable
```
private static final int FLAG_INIT_INCOMPLETE = FLAG_INIT | FLAG_INCOMPLETE;
```

Similar feedback for `FLAG_INIT_AND_FIN` below.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -3598,17 +3611,24 @@ private void onApplicationData(
...
+                        else if (deliveryId != abortedDeliveryId)
+                        {
+                            onApplicationDataContOrFin(traceId, reserved, authorization, flags, payload);
+                        }
                     }
-                    else if (deliveryId != abortedDeliveryId)
+                    else
                     {
-                        onApplicationDataContOrFin(traceId, reserved, authorization, flags, payload);
+                        flushReplySharedBudget(traceId);
```
jfallows: Suggest inverting this `if-else` such that the `if` condition handles `incomplete` as the special case, falling through to normal case in `else`.

It has the same efficiency at runtime, but makes it a bit easier to read the code.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -3234,7 +3242,11 @@ private void onDecodeTransfer(
                             this.messageFormat = messageFormat;
                         }
                     }
-                    if (!more)
+                    if (aborted)
+                    {
+                        flags = FLAG_INCOMPLETE;
+                    }
+                    if (!more && !aborted)
```
jfallows, line 3250: Suggest leaving a blank line between successive `if` blocks so they are not easily misread as `if-else`.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -1147,6 +1149,12 @@ private int decodeTransfer(
                 if (!sender.fragmented)
                 {
                     assert deliveryId != NO_DELIVERY_ID; // TODO: error
+                    if (transfer.hasAborted() && transfer.aborted() == 1)
+                    {
+                        progress = limit;
+                        server.decoder = decodePlainFrame;
+                        break decode;
+                    }
```
jfallows, line 1157: This is to handle the case where the first `transfer` fragment is `aborted` right away, so we ignore the message completely, right?

jfallows (changes requested):

jfallows (dismissed):

jfallows (approved):

Issue #41, opened 2020-09-11
Support pipelined OPEN


jfallows:
See AMQP 1.0 specification, [section 2.4.7](http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-transport-v1.0-os.html#doc-idp211568).

Proactively send `protocol-header` on connected, proactively send `OPEN` after `protocol-header` exchanged.

PR #89, opened 2020-10-14
Implement handle-max

### LINKED ISSUE #54: Reject `ATTACH` when `handle` exceeds `handle-max`
See AMQP 1.0 specification, [section 2.7.2](http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-transport-v1.0-os.html#type-begin).

izzymoriguchi:
Fix #54 
Requires https://github.com/reaktivity/nukleus-amqp.spec/pull/41

### DIFF
--- a/src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
+++ b/src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
@@ -2475,7 +2475,7 @@ public final class AmqpServerFactory implements StreamFactory
             decode:
             if (session != null)
             {
-                if (session.handleCount == decodeHandleMax)
+                if (attach.handle() > decodeHandleMax)
                 {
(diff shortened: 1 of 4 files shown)


src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -2843,6 +2862,7 @@ private void onDecodeAttach(
                         AmqpServerStream oldLink = links.put(handle, link);
                         assert oldLink == null;
                         link.onDecodeAttach(traceId, authorization, attach);
+                        handleCount++;
```
jfallows: Where is `handleCount--`?

jfallows (changes requested):

jfallows (dismissed):

jfallows (approved):

Issue #43, opened 2020-09-11
Verify simultaneous CLOSE


jfallows:
See AMQP 1.0 specification, [section 2.4.4](http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-transport-v1.0-os.html#doc-idp170544).

Ensure `CLOSE` frames can be sent simultaneously with different error codes.

PR #91, opened 2020-10-15
Support DETACH with session handle error case.

### LINKED ISSUE #55: Reject `ATTACH` when `handle` already in use
See AMQP 1.0 specification, [section 2.7.3](http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-transport-v1.0-os.html#type-attach).

izzymoriguchi:
Fix #55 
Requires https://github.com/reaktivity/nukleus-amqp.spec/pull/43


src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -2475,11 +2476,19 @@ private void onDecodeAttach(
...
             {
-                if (attach.handle() > decodeHandleMax)
+                final long handle = attach.handle();
+                if (handle > decodeHandleMax)
                 {
                     onDecodeError(traceId, authorization, CONNECTION_FRAMING_ERROR, null);
                     break decode;
                 }
+
+                final boolean handleInUse = session.links.values().stream().anyMatch(l -> l.handle == handle);
```
jfallows: Suggest we change this to
```
final boolean handleInUse = session.links.containsKey(handle);
```
for efficiency.

jfallows (changes requested):

jfallows (approved):

PR #92, opened 2020-10-16
Implement handling of exceeded max-message-size

### LINKED ISSUE #56: Reject `TRANSFER` when `max-message-size` exceeded
See AMQP 1.0 specification, [section 2.7.3](http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-transport-v1.0-os.html#type-attach).

### LINKED ISSUE #57: Send `aborted` `TRANSFER` to receiver when `max-message-size` exceeded
See AMQP 1.0 specification, [section 2.7.3](http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-transport-v1.0-os.html#type-attach).

>Any attempt to deliver a message larger than this results in a message-size-exceeded link-error.

Is it okay to treat this as a filter condition by message size, or must we also indicate to the client that they did not receive such a message?

When this occurs we should send an `aborted` `TRANSFER` so that the client is aware of the message and also that it is being skipped, in this case due to message size.

izzymoriguchi:
Fix #56 
Fix #57 
Requires https://github.com/reaktivity/nukleus-amqp.spec/pull/44

### DIFF
--- a/src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
+++ b/src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
@@ -3026,6 +3026,8 @@ public final class AmqpServerFactory implements StreamFactory
                 private StringFW addressFrom;
                 private StringFW addressTo;
                 private long decodeMaxMessageSize;
+                private long encodeMaxMessageSize;
+                private boolean maxMessageSizeExceeded;
 
                 private AmqpErrorType detachError;
 
@@ -3059,6 +3061,7 @@ public final class AmqpServerFactory implements StreamFactory
                 {
                     this.name = attach.name().asString();
                     this.handle = attach.handle();
+                    this.encodeMaxMessageSize = attach.hasMaxMessageSize() ? attach.maxMessageSize() : 0;
 
                     final AmqpCapabilities capability = amqpCapabilities(role);
                     final AmqpSenderSettleMode amqpSenderSettleMode = attach.sndSettleMode();
@@ -3157,7 +3160,7 @@ public final class AmqpServerFactory implements StreamFactory
                         }
                     }
 
-                    if (size > defaultMaxMessageSize)
+                    if (defaultMaxMessageSize > 0 && size > defaultMaxMessageSize)
                     {
                         onDecodeError(traceId, authorization, LINK_MESSAGE_SIZE_EXCEEDED);
                     }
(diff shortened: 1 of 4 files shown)


src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -3493,11 +3518,20 @@ private void onApplicationDataInit(
                         transferBuilder.more(1);
                     }
 
-                    final AmqpTransferFW transfer = transferBuilder.build();
                     final DirectBuffer fragmentBuffer = messageFragment.buffer();
                     final int fragmentOffset = messageFragment.offset();
                     final int fragmentLimit = messageFragment.limit();
                     final int fragmentSize = fragmentLimit - fragmentOffset;
+
+                    if (encodeMaxMessageSize > 0 && fragmentSize > encodeMaxMessageSize)
```
jfallows: `fragmentSize + deferred`?

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -3551,11 +3585,22 @@ private void onApplicationDataContOrFin(
                         transferBuilder.more(1);
                     }
 
-                    final AmqpTransferFW transfer = transferBuilder.build();
                     final DirectBuffer fragmentBuffer = messageFragment.buffer();
                     final int fragmentOffset = messageFragment.offset();
                     final int fragmentLimit = messageFragment.limit();
                     final int fragmentSize = fragmentLimit - fragmentOffset;
+                    if (maxMessageSizeExceeded)
```
jfallows: Likely we should avoid calling `onApplicationDataContOrFin` again for same deliveryId in case we get `DATA (init)`, `DATA (cont)`, `DATA (fin)` and the max message size is exceeded in `DATA (init)` or `DATA (cont)`.

This could be done from `onApplicationData`, and rather than tracking a `boolean` for `maxMessageSizeExceeded` we could consider adding `abortedDeliveryId` and assigning it to the current `deliveryId` at the time we send `TRANSFER aborted`.

Then in `onApplicationData` we would have something like:
```
                    if ((flags & FLAG_INIT) == FLAG_INIT)
                    {
                        deliveryId++;
                        onApplicationDataInit(traceId, reserved, authorization, flags, extension, payload);
                    }
                    else if (deliveryId != abortedDeliveryId)
                    {
                        onApplicationDataContOrFin(traceId, reserved, authorization, flags, payload);
                    }
```
which means we wouldn't need to reset `maxMessageSizeExceeded` back to false, as `deliveryId != abortedDeliveryId` will implicitly become true for the next delivery, as needed.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -3493,11 +3518,20 @@ private void onApplicationDataInit(
...
                     final int fragmentOffset = messageFragment.offset();
                     final int fragmentLimit = messageFragment.limit();
                     final int fragmentSize = fragmentLimit - fragmentOffset;
+
+                    if (encodeMaxMessageSize > 0 && fragmentSize > encodeMaxMessageSize)
+                    {
+                        transferBuilder.aborted(1);
+                        doEncodeTransfer(traceId, authorization, outgoingChannel, transferBuilder.build(),
+                            fragmentBuffer, fragmentOffset, 0);
+                        maxMessageSizeExceeded = more;
```
jfallows: With the changes proposed above, this simplifies
```
transferBuilder.aborted(1);
abortedDeliveryId = deliveryId;
fragmentSize = 0;
```

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -3551,11 +3585,22 @@ private void onApplicationDataContOrFin(
...
+                    if (maxMessageSizeExceeded)
+                    {
+                        transferBuilder.aborted(1);
+                        doEncodeTransfer(traceId, authorization, outgoingChannel, transferBuilder.build(),
+                            fragmentBuffer, fragmentOffset, 0);
+                        if (!more)
+                        {
+                            maxMessageSizeExceeded = false;
+                        }
+                        return;
```
jfallows: This code would simplify similarly as above, but we also need to detect when the aggregate size across all `TRANSFER` frames now exceeds the `max-message-size` in total.

jfallows (changes requested):

jfallows (dismissed):

jfallows (approved):

Issue #50, opened 2020-09-11
Support flow control echo from receiver


jfallows:
See AMQP 1.0 specification, [section 2.6.10](http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-transport-v1.0-os.html#doc-idp429232).

Stopping the transfers on a given link is accomplished by updating the link-credit to be zero and sending the updated flow state.

It is possible that some transfers could be in flight at the time the flow state is sent, so incoming transfers could still arrive on the link.

The echo field of the flow frame MAY be used to request the sender's flow state be echoed back.

This MAY be used to determine when the link has finally quiesced.

PR #98, opened 2020-10-21
Support sequence number

### LINKED ISSUE #64: Support `sequence-no` via RFC-1982 serial number
See AMQP 1.0 specification, [section 2.8.10](http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-transport-v1.0-os.html#type-sequence-no).

Specifically, handle case where any `sequence-no` type field is approaching maximum value and needs to wrap.

izzymoriguchi:
Fix #64 
Requires https://github.com/reaktivity/nukleus-amqp.spec/pull/50

### DIFF
--- a/src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
+++ b/src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
@@ -1141,15 +1141,14 @@ public final class AmqpServerFactory implements StreamFactory
                 if (!sender.fragmented)
                 {
                     assert deliveryId != NO_DELIVERY_ID; // TODO: error
-                    final long previousDeliveryId = (session.remoteDeliveryId + 1) & 0xFFFF_FFFFL;
-                    if (deliveryId != previousDeliveryId)
+                    session.remoteDeliveryId = (session.remoteDeliveryId + 1) & 0xFFFF_FFFFL;
+                    if (deliveryId != session.remoteDeliveryId)
                     {
                         server.onDecodeError(traceId, authorization, INVALID_FIELD, null);
                         progress = limit;
                         server.decoder = decodePlainFrame;
                         break decode;
                     }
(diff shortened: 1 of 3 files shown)


src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -1143,7 +1141,14 @@ private int decodeTransfer(
                 if (!sender.fragmented)
                 {
                     assert deliveryId != NO_DELIVERY_ID; // TODO: error
-                    assert deliveryId == session.remoteDeliveryId + 1; // TODO: error
+                    final long previousDeliveryId = (session.remoteDeliveryId + 1) & 0xFFFF_FFFFL;
```
jfallows: Suggest renaming `previousDeliveryId` to `nextDeliveryId`.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -1143,7 +1141,14 @@ private int decodeTransfer(
...
+                    if (deliveryId != previousDeliveryId)
+                    {
+                        server.onDecodeError(traceId, authorization, INVALID_FIELD, null);
+                        progress = limit;
+                        server.decoder = decodePlainFrame;
+                        break decode;
+                    }
```
jfallows, line 1151: Alternatively, advance session remote delivery id proactively, then check for correctness and close with error if it doesn't line up.
```
session.remoteDeliveryId = (session.remoteDeliveryId + 1) & 0xFFFF_FFFFL;
if (deliveryId != session.remoteDeliveryId)
{
   ...error...
}
```

jfallows (changes requested):

jfallows (dismissed):

jfallows (approved):

Issue #53, opened 2020-09-11
Process `OPEN` with `outgoing-locales` when `en-US` is omitted


jfallows:
See AMQP 1.0 specification, [section 2.7.1](http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-transport-v1.0-os.html#type-open).

PR #101, opened 2020-11-04
Support application-properties with amqp simple types

### LINKED ISSUE #70: Reject `application-properties` map entries with invalid key or value type
See AMQP 1.0 specification, [section 3.2.5](http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-messaging-v1.0-os.html#type-application-properties).

izzymoriguchi:
Fix #70 
Requires https://github.com/reaktivity/nukleus-amqp.spec/pull/53

### DIFF
--- a/src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
+++ b/src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
@@ -115,6 +115,7 @@ import org.reaktivity.nukleus.amqp.internal.types.Flyweight;
 import org.reaktivity.nukleus.amqp.internal.types.OctetsFW;
 import org.reaktivity.nukleus.amqp.internal.types.String8FW;
 import org.reaktivity.nukleus.amqp.internal.types.StringFW;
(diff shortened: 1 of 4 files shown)


src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -4153,17 +4158,14 @@ private void encodeApplicationProperties(
...
             AmqpApplicationPropertyFW item)
         {
-            int  valueOffset = 0;
-            AmqpStringFW key = amqpStringRW.wrap(valueBuffer, valueOffset, valueBuffer.capacity())
-                .set(item.key())
-                .build();
-            valueOffset += key.sizeof();
+            DirectBuffer buffer = item.value().bytes().buffer();
+            int offset = item.value().bytes().offset();
+            int limit = item.value().bytes().limit();
```
jfallows: Recommend we avoid calling `item.value().bytes()` repeatedly.

jfallows (changes requested):

jfallows (dismissed):

jfallows (approved):

PR #103, opened 2020-12-18
Support flow control acknowledgement

izzymoriguchi:
Requires https://github.com/reaktivity/nukleus-amqp.spec/pull/55


src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -3509,11 +3676,11 @@ private void flushInitialWindow(
                 {
                     if (AmqpState.replyOpened(state) && role == SENDER)
                     {
-                        this.remoteLinkCredit = (int) (Math.min(bufferPool.slotCapacity(), initialBudget) /
+                        this.remoteLinkCredit = (int) (Math.min(bufferPool.slotCapacity(), initialMax) /
                                                        Math.min(bufferPool.slotCapacity(), decodeMaxFrameSize));
-                        maximum.value = 0;
-                        links.values().forEach(l -> maximum.value += l.remoteLinkCredit);
-                        incomingWindow = maximum.value;
+                        max.value = 0;
```
jfallows: Suggest renaming `max` to `remoteLinkCreditSum`.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -2388,31 +2534,29 @@ private void doNetworkReset(
...
 
         private void decodeNetworkIfNecessary(
             long traceId)
         {
             if (decodeSlot != NO_SLOT)
             {
-                final long authorization = 0L; // TODO
-                final long budgetId = 0L; // TODO
+                final long authorization = 0L;
+                final long budgetId = 0L;
```
jfallows: I think we should keep the `TODOs` here as the rationale for having them has not changed.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -2173,20 +2297,30 @@ private void onNetworkAbort(
...
+            replyPad = padding;
 
-            replyBudget += credit;
-            replyPadding = padding;
+            assert replyAck <= replySeq;
 
             if (replyBudgetReserved > 0)
             {
-                final int reservedCredit = Math.min(credit, replyBudgetReserved);
+                final int reservedCredit = Math.min(replyMax, replyBudgetReserved);
```
jfallows: Changing `credit` to `replyMax` changes the intended logic for `reservedBudgetReserved`, agreed?

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -2091,24 +2167,42 @@ private void onNetwork(
...
+            assert acknowledge <= sequence;
+            assert sequence >= initialSeq;
+            assert acknowledge <= initialAck;
+
+            initialSeq = sequence + data.reserved();
 
-            if (initialBudget < 0)
+            assert initialAck <= initialSeq;
+
+            if (initialSeq > initialAck + bufferPool.slotCapacity())
```
jfallows: Should this be `if (initialSeq > initialAck + initialMax)`?

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -2388,31 +2534,29 @@ private void doNetworkReset(
...
+            int maximum,
             long budgetId)
         {
-            assert credit > 0;
-
             state = AmqpState.openInitial(state);
 
-            initialBudget += credit;
-            doWindow(network, routeId, initialId, traceId, authorization, budgetId, credit, padding, 0);
+            doWindow(network, routeId, initialId,  initialSeq, initialAck, maximum, traceId, authorization, budgetId, padding, 0);
```
jfallows: Please see https://github.com/reaktivity/nukleus-kafka.java/blob/feature/flow-control-ack/src/main/java/org/reaktivity/nukleus/kafka/internal/stream/KafkaClientProduceFactory.java#L1035 for an example of modeling `doXXXWindow` for initial stream.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -2139,6 +2233,36 @@ private void onNetworkData(
...
+
+            assert acknowledge <= sequence;
+            assert sequence >= initialSeq;
+            assert acknowledge <= initialAck;
+
+            initialSeq = sequence + flush.reserved();
+
+            assert initialAck <= initialSeq;
+
+            if (initialSeq > initialAck + bufferPool.slotCapacity())
```
jfallows: Should be `if (initialSeq > initialAck + initialMax)`?

jfallows (changes requested):

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -2543,20 +2548,32 @@ private void doNetworkWindow(
...
+                initialAck = newInitialAck;
+                assert initialAck <= initialSeq;
+
+                initialMax = minInitialMax;
+            }
+
             state = AmqpState.openInitial(state);
 
-            doWindow(network, routeId, initialId,  initialSeq, initialAck, maximum, traceId, authorization, budgetId, padding, 0);
+            doWindow(network, routeId, initialId, initialSeq, initialAck, maximum, traceId, authorization, budgetId, padding, 0);
```
jfallows: Please pass `initialMax` for `maximum` and remove `maximum` from the `doNetworkWindow` parameter list as we don't need it any more.

src/main/java/org/reaktivity/nukleus/amqp/internal/stream/AmqpServerFactory.java
```diff
@@ -2568,7 +2585,7 @@ private void decodeNetworkIfNecessary(
                 final int initialCredit = reserved - decodeSlotReserved;
                 if (initialCredit > 0)
                 {
-                    doNetworkWindow(traceId, authorization, initialCredit, 0, 0);
+                    doNetworkWindow(traceId, authorization, initialCredit, 0, 0, decodeSlotOffset, initialMax);
```
jfallows: We are passing `initialCredit` for `padding` here, likely it can be removed now that `maximum` parameter is gone and pass `0` for `padding` instead (as was the original intent, i think).

However, the logic of this still needs to be addressed given that `initialCredit` would no longer be used in the method call to `doNetworkWindow`.

jfallows (changes requested):

jfallows (approved):

